PR #85, opened 2021-06-01
Add fancy ascii header

ewels:
Because no command line tool is complete without an ascii header 😆

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -49,6 +49,19 @@ console = rich.console.Console()
 ###############################################################################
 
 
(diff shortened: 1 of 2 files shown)


ewels:
ps. I challenge others to come up with nicer artwork for the logo! I had a quick play with online generator things but couldn't get any results that I was happy with at such a small size, so just ended up doing it by hand. Maybe others with more patience than me might be able to do a better version though..

inaod568:
I like it! 😄

inaod568 (approved):

Issue #78, opened 2021-05-28
Python best practices - system exits & logging


ewels:
## System exits

Doing a hard system exit is an absolute last resort. Normally you want to be raising an exception instead, then probably having a single location right at the top of the code to catch this and do the system exit with non-zero code.

This is especially important if other code packages are importing and using the functions. System exits cannot be caught, so for example if used in the `dds_web` code then these will crash the server. Exceptions can be caught and handled differently depending on where the function is being used.

In most cases here I think it will make sense to create your own exception types. You can then pass the error message in the `raise` statement and log it before exiting downstream (eg. [as done here](https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/nf_core/__main__.py#L251-L253)).

I use this pattern a lot in nf-core and typically do the exit call in the command-line handling code. eg. [here](https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/nf_core/__main__.py#L520-L522). So then if you have 5 subcommands you'll probably have a maximum of 5 exit calls. You can be pretty sure that no-one else will be importing and reusing your cli handling code.

In some cases you have an exit code `0` because it's something like just not having any files to show etc - in other words, normal behaviour. Here you should probably just use `return` to drop out of the function execution early without an exception.

## Logging

You almost never want to print to the console using a `rich` `Console`.

One reason is that this prints to standard out - but most of the current usage is log / status messages. Normally, these should be going to standard error and only the real "results" (eg. the list of projects etc) should go to standard out. This means that command line users can split the output types for downstream use.

Much like the system exits / exceptions, the console calls are not useful for other tools importing the functions. It's better to use the `logging` library instead - then the log messages can be assigned to a namespace and their output customised by any tool using the function (eg. [only showing errors](https://github.com/nf-core/tools/blob/7721f3742fc30a9e6e46d879404bd1366345092d/nf_core/lint/actions_schema_validation.py#L25-L26) or being channeled to a web server log).

[...]

PR #86, opened 2021-06-02
System exit cleanup

ewels:
WIP PR as I try to address #78.

I'm basically doing three things as I go through each command:

* Move as much code out of `__main__.py` as possible into the module files, under a new function call
* Raise new custom exception types in case of errors instead of calling system exits
* Catch those exception types in `__main__.py` and log the messages, call a system exit.

The logic is that basically no other tool will ever import `__main__.py` so it's fine to put command-line specific code there such as `click` and hard exit calls. All of the main functionality goes into modules that can be imported and reused by other code bases, using exceptions that can be caught and dealt with differently elsewhere if needed.

While I'm going through, I will try to convert rich console print statements to log calls where I see them.


ewels:
So that this refactoring doesn't get too huge, I will continue the other parts in separate branches and wait for this to be merged before going any further. Then I will do one PR per section.

So feel free to review + test 😃

ewels:
Ok I lied, I did `dds ls` too. Bit it was a small one. Bed now.. 😅

ewels:
I noticed that the package version was defined in two locations - fixed in 0f9af86 so that there is just one: in `setup.py`.

ewels:
Ok, _really_ done with this PR now.. 😅 

Will branch before I touch anything else (barring PR review comments).

dds_cli/__main__.py
```diff
@@ -201,117 +195,22 @@ def put(
...
+            username,
+            project,
+            source,
+            source_path_file,
+            break_on_fail,
+            overwrite,
+            num_threads,
+            silent,
+        )
+    except (dds_cli.exceptions.AuthenticationError, dds_cli.exceptions.UploadError) as e:
```
inaod568, line 211: I get the following error: 
```
File "/Users/inaod568/repos/DDS/dds_cli/dds_cli/exceptions.py", line 11, in __str__
    return f"\n:no_entry_sign: {self.message} :no_entry_sign:\n"
AttributeError: 'AuthenticationError' object has no attribute 'message'
```

Don't have time to fix it now though, sorry

ewels: How did you manage to trigger this? The place it is raised is in the `if self.method != "put":` check and I don't know how to make that fail.. 😅  (so struggling to replicate to fix)

inaod568 (approved):

PR #88, opened 2021-06-04
`dds ls` usability improvements

ewels:
Work in progress, trying to make the `dds ls` command easier to use (from my perspective) and a bit fancier.

Will add to this list as I go and mark as ready to review when I think the PR has enough :)

* Clean up command line flags
    * Positional only for project and folder (reversed order)
    * `--projects` removed and inferred by lack of specified `project`
    * More cli help text
* Instead of warning if there is a lot of output, automatically page the output
    * Use the [`rich` console pager](https://rich.readthedocs.io/en/stable/console.html#paging)
    * Base the decision on the height of the user's terminal, not an arbitrary count
    * Non-interactive usage (eg. piping / redirecting to a file) will still have normal behaviour
* Prompt for a project to view files for
    * After printing table of projects, use [Questionary to prompt](https://questionary.readthedocs.io/en/stable/pages/types.html#autocomplete) for a project ID to show the files for
    * Only prompts if the console is running interactively (eg. not redirection)
    * Live validation of entered text against possible project IDs
* Prompt for a folder to view files inside
    * After listing files, prompt for a subfolder to view
    * Only prompts if console is interactive
    * Only prompts if there is a folder in the list
    * _Note: It would be nicer to be able to use the [Questionary File Path](https://questionary.readthedocs.io/en/stable/pages/types.html#file-path) prompt and to be able to tab through subdirectories. Currently the website does not return a recursive file listing though, so this isn't possible._

### DIFF
--- a/dds_cli/__init__.py
+++ b/dds_cli/__init__.py
@@ -2,6 +2,7 @@
 
 import os
 import pkg_resources
+import prompt_toolkit
 
 ###############################################################################
 # PROJECT SPEC ################################################# PROJECT SPEC #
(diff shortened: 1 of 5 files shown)


ewels:
Ok, I was having fun and suddenly it's now late and I finished the list of things I wanted to do for this PR in one sitting 👀

ewels:
Reminder from discussion with @inaod568 on Slack - the change in command line handling here might break things for users who define a project in a DDS config file.

My preference would be to remove the ability to set project in a config file. If we're going for that, need to do so in this PR I guess.

inaod568:
Looks good and I agree with the changes. Just discussing the point about removing the `project` field from the config file. Will merge after that.

inaod568:
Decisions made in DC group: 
- Listing projects: both options
    - `--projects` --> `dds ls --projects`
    - no specified project --> `dds ls`
- Project ID only specified with `--project/-p`, removing field from config file
- `--project/-p` option for `dds ls` kept - as little difference between the different commands as possible

ewels:
Ok 👍🏻 So basically just add back `--projects`?

inaod568:
Yes!

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

inaod568 (approved):

ewels:
@inaod568 I guess you need to clean up this parameter too?

https://github.com/ScilifelabDataCentre/dds_cli/blob/86f0ff444a007607cbd9013a07cea3b4b16a03e0/dds_cli/base.py#L66

inaod568:
Yes, will do 👍

Den mån 14 juni 2021 kl 08:00 skrev Phil Ewels ***@***.***>:

Issue #129, opened 2021-06-29
Fix glitchy progress bars


ewels:
Since I messed around with the logging code, the progress bar used for uploads and downloads now jumps whenever a log message is printed, instead of elegantly dropping down below each new message.

This is because the log handler and the progress bar are using different rich `Console` objects. It should be possible to fix by sharing the same `Console` between both. See https://github.com/willmcgugan/rich/discussions/1317#discussioncomment-938183 for an example.

I think that to do this, it maybe makes sense to create a `Console` object that can be returned from a `utils` module or something for easy reused between disparate parts of the codebase. But whatever makes sense really.

PR #130, opened 2021-08-11
Fix glitchy progress bars

ErikDanielsson:
Fixed glitchy  progress bars. This was done in the way that was suggested in #129 -- by creating a `Console` object in `utils.py` which is then reused by all files writing output and by the progress bars.


inaod568 (approved):
Snyggt!

PR #151, opened 2021-09-09
Closes #148

### LINKED ISSUE #148: DDS_METHODS not used?
https://github.com/ScilifelabDataCentre/dds_cli/blob/7b09e75886ee96391b15f08163443280f4b4ee9a/dds_cli/__init__.py#L20

Just doing a bit of code review to get into the code. I saw this was added recently in https://github.com/ScilifelabDataCentre/dds_cli/commit/1d99e6fda0ead9c706b90e4e938da51e68ca9946, but it doesn't seem to be used anywhere? I'm not sure it's needed either, I guess click should already keep track of which commands are allowed? But maybe I'm wrong?

inaod568:


inaod568:
Merging this despite failed DDS pytest because complaining on a bit that should probably be changed.

PR #159, opened 2021-09-22
Include legacy detection

### LINKED ISSUE #93: Fix Windows icon compatibility
The CLI only displays small question mark symbols instead of up/down/lock/warning symbols etc.

MatthiasZepper:
Addresses issue #93. 

Suggested changes:

- Introduce a new variable `dds_on_legacy_console` in the `__init__.py`  that stores the return value of `rich.console.detect_legacy_windows()`. Otherwise the detection would need to occur every time the method `TextHandler.task_name` is called.
- This variable is imported into `text_handler.py` and a different dictionary is used to resolve the `step`, which contains spelled out status messages instead of emojis. 

Testing:

- Tested to the extent that the dds command doesn't crash right away.
- `rich.console.detect_legacy_windows()` returns TRUE on a Windows 8 cmd.exe (tested before with a separate example script).
- So far, I have not tested if and how the alternative messages are displayed on a legacy console as that would require the whole Docker dds_web backend to be run inside the Virtual Machine. Will wait to test this until we have a public API up and running on the web.

Subject to further discussion:

- Naming of the alternative status messages: _encrypt_ and _decyrpt_ were replaced by _seal_ and _open_ respectively to shorten both to 4 letters. Alternatively, something like _ecpt/dcpt_ or _ecph/dchp_ (encipher/decipher) could be used.
- Formatting of the status messages: Bold or anything else like a specific color?

For the record:

-  `rich.console.detect_legacy_windows()` is unfortunately only a proxy to detect if emojis will be displayed correctly. [Unicode support is mandatory and font fallback required](https://devblogs.microsoft.com/commandline/windows-command-line-unicode-and-utf-8-output-text-buffer/), if the chosen font doesn't have glyphs for the respective emojis. Windows legacy consoles never support font fallback and rarely Unicode. Font fallback might be an issue on other terminal emulators and operating systems, too and is not tested for with above condition.
- Emojis are also used at other parts of the code, e.g. in the `exceptions.py`. Currently they are not stripped from the output there, so the current pull request does not fully resolve #93 .


alneberg:
I think the changes in `.gitignore` would be fine to keep for everyone so don't bother changing those. 👍

dds_cli/text_handler.py
```diff
@@ -7,6 +7,7 @@
 # Standard library
 import logging
 import pathlib
+from dds_cli import dds_on_legacy_console
```
inaod568, line 10: Change to `import dds_cli` and use as `dds_cli.dds_on_legacy_console` instead. 
Also add `dds_on_legacy_console` to the `__all__` list in `__init__.py`

inaod568 (changes requested):

inaod568:
Other than the requested change it looks good!

dds_cli/text_handler.py
MatthiasZepper: Thanks for the fast review. I changed what you requested. On the current dev branch we don't have an `_all_` list yet and also on the other branches that I checked (such as your dev) I couldn't find one.  So I created that list with all the symbols included. I hope this was what you intended to do.

inaod568:
Run `black .` in your `dds_cli` folder, or if you have VSCode you can also set black to be automatically run on save

dds_cli/text_handler.py
inaod568: Yes `__all__ = ['DDSEndpoint','FileSegment','dds_questionary_styles', 'dds_on_legacy_console']` was exactly what I meant!

MatthiasZepper:
> @MatthiasZepper Run `black .` in your `dds_cli` folder, or if you have VSCode you can also set black to be automatically run on save

@inaod568 Beyond comprehension: Configured everything together with Johannes last week, so that black runs automatically with every time a file is saved in VS Code. This also worked on last Thursday and Friday. Apparently MacOS update also broke that and I didn't notice, because the other stuff in VSCode like the extensions still worked like a charm.

dds_cli/text_handler.py
```diff
@@ -65,13 +66,18 @@ def task_name(file, step="", max_len=30):
...
-        elif step == "put":
-            symbol = ":arrow_up: "
-        elif step == "get":
-            symbol = ":arrow_down: "
-        elif step == "decrypt":
-            symbol = ":unlock:"
-        return f"{symbol} {task_name}"
+        if dds_cli.dds_on_legacy_console:
+            symbols = {
+                "encrypt": "[bold]seal",
```
inaod568, line 71: Small thing here but I think it's better to have `encrypt` and `decrypt` instead of `seal` and `open`, respectively

inaod568 (changes requested):

inaod568:
> > @MatthiasZepper Run `black .` in your `dds_cli` folder, or if you have VSCode you can also set black to be automatically run on save
> 
> @inaod568 Beyond comprehension: Configured everything together with Johannes last week, so that black runs automatically with every time a file is saved in VS Code. This also worked on last Thursday and Friday. Apparently MacOS update also broke that and I didn't notice, because the other stuff in VSCode like the extensions still worked like a charm.

Ah yeah, could also be a VSCode thing. It does weird things some times. But great that it's working again / that black has been run again 👍🏻

dds_cli/text_handler.py
MatthiasZepper: So you have no concerns regarding long status codes of variable length? Just asking, because I believed it might affect the columnar arrangement and ultimately readability when working with many files. I noticed you specifically created a function to truncate long filenames and believed it was for this reason...

inaod568 (approved):
`rich` handles the display witdh etc automatically, so the extra 3 characters here really shouldn't affect this. The only thing that will happen is that the actual progress bar will become shorter. So the "todo test for display width" thing is not something we need to put a lot of effort into, this should be a sufficient solution 👍🏻

PR #171, opened 2021-10-11
Add sensitive option in project creation

aanil:


alneberg (approved):
Looks good and seems to work!

PR #179, opened 2021-10-21
Some refactoring of `dds rm` and config finding/parsing errors

alneberg:
A bit of a mixture in here, just things I found could use some improvement along with removing file and line origin of log statements by default. 

 - Improved but simplified handling of errors related to config finding
 - Show line numbers for log entries only if `--verbose` is used
 - Raise exceptions instead of returning messages for `dds rm` errors
 - Made the failure report table an attribute instead of a message
 - Replaced a bunch of `os._exit` with `sys.exit`

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -79,6 +79,7 @@ def dds_main(ctx, verbose, log_file):
                 console=dds_cli.utils.console,
                 show_time=False,
                 markup=True,
+                show_path=verbose,
             )
         )
 
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -385,12 +385,12 @@ def rm(dds_info, proj_arg, project, username, rm_all, file, folder, config):
     # One of proj_arg or project is required
     if all(x is None for x in [proj_arg, project]):
         LOG.error("No project specified, cannot remove anything.")
-        os._exit(1)
+        sys.exit(1)
 
     # Either all or a file
     if rm_all and (file or folder):
         LOG.error("The options '--rm-all' and '--file'/'--folder' cannot be used together.")
-        os._exit(1)
+        sys.exit(1)
 
     project = proj_arg if proj_arg is not None else project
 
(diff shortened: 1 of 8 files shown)


dds_cli/__main__.py
```diff
@@ -97,7 +98,10 @@ def dds_main(ctx, verbose, log_file):
         if not any(x in sys.argv for x in ["--config", "-c", "--username", "-u"]):
             config_file = pathlib.Path().home() / pathlib.Path(".dds-cli.json")
             if not config_file.is_file():
-                raise exc.ConfigFileNotFoundError(config_file)
+                LOG.error(
+                    f"Username omitted and config file not specified nor found at default path: {config_file}"
+                )
+                sys.exit(1)
```
inaod568, line 104: I thought that `sys.exit(1)` would give a really ugly message but it appears it doesn't. So 👍🏻

dds_cli/cli_decorators.py
```diff
@@ -191,43 +191,35 @@ def create_and_remove_task(self, *args, **kwargs):
...
             elif func.__name__ == "remove_folder":
-                description = "Removing folder(s)..."
+                description = "Removing folder(s)"
 
             # Add progress task
-            task = progress.add_task(description=description)
+            task = progress.add_task(description=f"{description}...")
 
-            # Execute function
-            message = func(self, *args, **kwargs)
```
inaod568, line 204: Is the progress task removed in this case? So that the spinner is not displayed if it fails?

inaod568 (commented):
Nice, thanks for cleaning all that up.

inaod568:
Also there are some unused imports, e.g. `requests` etc in the  `__main__.py`

dds_cli/cli_decorators.py
inaod568: Tried it out now and it doesn't, when the message was caught and then the `progress.remove_task` below was used, the spinner line was removed. 
With `message = func(self, *args, **kwargs)` and `progress.remove_task(task)`:
```
Failed to delete files in project: {
    "message": "The project is empty.",
    "status": 400
}
```

With `func(self, *args, **kwargs)`:
```
Removing all files in project project_1... ⠀⠠
ERROR    Failed to delete files in project: {                                                                                                                      
             "message": "The project is empty.",                                                                                                                   
             "status": 400                                                                                                                                         
         }                                                                                                                                                         
             
```

PR #186, opened 2021-11-02
Add functionality to associate users with projects

aanil:


inaod568 (approved):
Looks good and I can't find the issue I found previously. 
A couple of things I noticed / thought of though (which are not neccessarily connected to the CLI but just this functionality in general): 
- `dds create`: At the moment the same user can be specified in the `--owner` and `--researcher` 
- `dds add-user`: If the user doesn't exist then it says "Invite successful" but at the moment the user is only invited to the dds in general and not the project so perhaps we should add that to the message - that they need to add the user again when they have an account or something

aanil:
* Added a check for duplicates in owners and researchers
* Added a message clarifying that invited users need to be added to a project once they have an account.

PR #190, opened 2021-11-05
Remove config option

inaod568:
Removed config option, set username as required and removed the default value None for username. Also cleaned up some docstrings etc.


alneberg (approved):
Looks great! The test placeholder needed a special mention, but its good to keep it there.

Issue #187, opened 2021-11-02
Possibility to save and reuse a token


alneberg:
I would suggest a new endpoint for example:

```
dds login
```
or perhaps
```
dds session
```
That would fetch a fresh token and either return it on the command line or save it to a file. 

If such token exists/is given it should be used for all requests until it fails or is renewed.

This relates to the possibility of using the cli in an automated fashion.

inaod568:
From discussion with @alneberg: `--username` does not override the use of the token. If it's another user, it's the users responsibility to have a logout routine and to use the correct account when using DDS. A possible future enhancement can be to add a `--force-auth` flag to not use the saved token, however this is IF the users mention that this is a useful feature.

PR #193, opened 2021-11-16
Save and reuse encrypted token

alneberg:
Added the possibility of saving and reusing the encrypted token so that users are not always prompted for passwords.

Added two new commands:
```bash
dds session -u username
```
to force renew the token and
```bash
dds session -u username --check
```
to check the token based on its age.

I modified the base handler a bit for the `--check` to not authenticate against the api. 

I introduced three new variables:
TOKEN_FILE, TOKEN_MAX_AGE and TOKEN_WARNING_AGE. If the expiration time of tokens are changed on the api, the last two needs to be updated and users need to reinstall the new version of the cli package.

solves issue #187


dds_cli/user.py
```diff
@@ -77,4 +110,135 @@ def __authenticate_user(self, password):
...
+        Returns None if no valid token can be found."""
+
+        if not self.token_file.is_file():
+            LOG.debug(f"Token file {self.token_file} does not exist.")
+            return None
+
+        self.check_token_file_permissions()
+
+        if self.token_expired():
+            LOG.debug(f"No token retrieved from file, will fetch new token from api")
```
inaod568, line 134: There are a couple of f-strings are not using any variables.

inaod568 (approved):
A small comment but no dealbreaker - looks and works great!!

PR #195, opened 2021-11-17
Username not required

alneberg:
- Made username argument not required as this is only required when the session token needs to be renewed. 
- Instead made project argument required where it actually is, which removed the need for the __verify_input method.
- Got rid of the `MissingCredentialsException` class as the `AuthenticationError` is sufficient.

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -470,7 +470,7 @@ def rm(_, proj_arg, project, username, rm_all, file, folder):
     "-p",
     required=True,
     type=str,
-    help="Project ID to which you're uploading data.",
+    help="Project ID from which you're downloading data.",
 )
 @click.option(
     "--get-all",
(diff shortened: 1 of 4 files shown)


dds_cli/__main__.py
```diff
@@ -461,14 +461,14 @@ def rm(_, proj_arg, project, username, rm_all, file, folder):
...
     type=str,
     help="Your Data Delivery System username.",
 )
 @click.option(
     "--project",
     "-p",
-    required=False,
+    required=True,
     type=str,
     help="Project ID to which you're uploading data.",
```
inaod568, line 473: Can you also change this: `to` --> `from` and `uploading` --> `downloading`

inaod568 (approved):
Looks good!

dds_cli/__main__.py
alneberg: Fixed!

PR #196, opened 2021-11-17
Replacing bucket filename with UUIDs generated from the filename

MatthiasZepper:
This PR addresses [issue 689](https://github.com/ScilifelabDataCentre/dds_web/issues/689) from the web. Instead of the clear text filename, now a UUID is generated from it. 

The function additionally introduces a length check, because keys to S3 objects may not exceed 1024 bytes in UTF-8 encoding. UUIDs are by definition 16 bytes long, however, the folder and the public project ID, which are also part of the key may cause issues under rare circumstances.

Many umlauts/mutated vowels or other less frequent characters take up two bytes each, in MacOS even 3, so if the full path would be in the range of 300 to 500 characters, we might have in issue. Currently, the bucket name is just trimmed then, but an raising an exception might also be possible.  

_(corresponding database column is db.text, so this is no problem)_

### DIFF
--- a/dds_cli/file_handler_local.py
+++ b/dds_cli/file_handler_local.py
@@ -81,13 +81,21 @@ class LocalFileHandler(fh.FileHandler):
 
     # Static methods ############## Static methods #
     @staticmethod
-    def generate_bucket_filepath(filename="", folder=pathlib.Path("")):
+    def generate_bucket_filepath(filename="", folder=pathlib.Path(""), project=""):
         """Generates filename and new path which the file will be
         called in the bucket."""
 
         # Generate new file name
-        new_name = "".join([str(uuid.uuid4().hex[:6]), "_", filename])
-        return str(folder / pathlib.Path(new_name))
+        new_name = (
+            project
+            + "/"
+            + str(folder / pathlib.Path(str(uuid.uuid5(uuid.NAMESPACE_X500, filename))))
+        )
+        # max length of S3 key is 1024 bytes in UTF-8
+        # because UTF-8 is a variable length encoding, some characters take up 4 bytes.
+        # Umlauts like ö or å will be encoded to 2 bytes each in Windows/Linux but to 3 bytes each on Mac,
+        # because Windows/Linux use Normal-Form-Composed (NFC) Unicode, MacOS uses Normal-Form-Decomposed (NFD) Unicode.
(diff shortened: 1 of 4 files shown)


dds_cli/file_handler_local.py
```diff
@@ -81,13 +81,21 @@ def __init__(self, user_input, temporary_destination, project):
...
         called in the bucket."""
 
         # Generate new file name
-        new_name = "".join([str(uuid.uuid4().hex[:6]), "_", filename])
-        return str(folder / pathlib.Path(new_name))
+        new_name = (
+            project
+            + "/"
+            + str(folder / pathlib.Path(str(uuid.uuid5(uuid.NAMESPACE_X500, filename))))
+        )
```
inaod568, line 93: It looks like the project is a part of the path now, but since the bucket is project specific this is not needed. Also, since the project may produce length errors I'd just skip that part. Just have 
```
        new_name = (
            str(folder / pathlib.Path(str(uuid.uuid5(uuid.NAMESPACE_X500, filename))))
        )
```
This also matches the old structure so there's less risk that other stuff is affected.

inaod568 (approved):

inaod568 (changes requested):
It looks like the project is a part of the path now, but since the bucket is project specific this is not needed. Also, since the project may produce length errors I'd just skip that part. Just have 
```
        new_name = (
            str(folder / pathlib.Path(str(uuid.uuid5(uuid.NAMESPACE_X500, filename))))
        )
```
This also matches the old structure so there's less risk that other stuff is affected.

dds_cli/file_handler_local.py
```diff
@@ -121,25 +129,25 @@ def __collect_file_info_local(self, all_paths, folder=pathlib.Path(""), task_nam
...
+                    "size_raw": path.stat().st_size,
+                    "compressed": is_compressed,
+                    "path_processed": path_processed,
+                    "size_processed": 0,
+                    "path_remote": self.generate_bucket_filepath(
+                        filename=path_processed.name, folder=folder
+                    ),
+                    "overwrite": False,
+                    "checksum": "",
+                }
```
inaod568, line 150: Unindenting here should be fine 👍🏻

inaod568 (approved):
Looks great!

PR #198, opened 2021-11-22
No Prompt and more

alneberg:
- Implemented the `--no-prompt` flag to fail rather than ask for missing arguments.
- Renamed Session to Auth everywhere
- New subcommands: `dds auth login`, `dds auth logout` and `dds auth info`.
- Removed some demo data.
- Tiny improvements on printing logic

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -14,7 +14,6 @@ import sys
 # Installed
 import click
 import click_pathlib
-from pkg_resources import parse_requirements
 import rich
 import rich.logging
 import rich.progress
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -813,6 +813,7 @@ def logout():
     try:
         with dds_cli.auth.Auth(username=None, authenticate=False) as authenticator:
             authenticator.logout()
+            LOG.info(f"[green]Successfully logged out![/green]")
     except dds_cli.exceptions.DDSCLIException as err:
         LOG.error(err)
         sys.exit(1)
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -809,7 +809,22 @@ def login(click_ctx, username):
 
 
 @auth.command()
+def logout():
+    """Remove the saved authentication token by deleting the '.dds_cli_token' file."""
+    try:
+        with dds_cli.auth.Auth(username=None, authenticate=False) as authenticator:
+            authenticator.logout()
+    except dds_cli.exceptions.DDSCLIException as err:
+        LOG.error(err)
+        sys.exit(1)
+
+
+@auth.command()
 def info():
     """Print info on saved authentication token validity and age."""
-    with dds_cli.auth.Auth(username=None, check=True) as authenticator:
-        authenticator.check()
+    try:
+        with dds_cli.auth.Auth(username=None, authenticate=False) as authenticator:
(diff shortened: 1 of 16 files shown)


dds_cli/auth.py
```diff
@@ -21,24 +17,34 @@
...
+            token_file.token_report()
+        else:
+            LOG.error(f"[red]No saved authentication token found![/red]")
+
+    def logout(self):
+        token_file = user.TokenFile()
+        if token_file.file_exists():
+            token_file.delete_token()
+        else:
+            LOG.info(f"[green]Already logged out![/green]")
```
inaod568, line 50: Nice 👍🏻

dds_cli/user.py
```diff
@@ -156,6 +163,10 @@ def read_token(self):
         LOG.debug("Token retrieved from file.")
         return token
 
+    def file_exists(self):
+        """Returns True if the token file exists."""
+        return self.token_file.is_file()
```
inaod568, line 168: Since the  `token_file` is a `Path` I'm thinking we should just use `self.token_file.exists()` instead of the `def file_exists()` function?

dds_cli/__main__.py
```diff
@@ -742,40 +768,62 @@ def create(
...
+
+@auth.command()
+def logout():
+    """Remove the saved authentication token by deleting the '.dds_cli_token' file."""
+    try:
+        with dds_cli.auth.Auth(username=None, authenticate=False) as authenticator:
+            authenticator.logout()
+    except dds_cli.exceptions.DDSCLIException as err:
+        LOG.error(err)
+        sys.exit(1)
```
inaod568, line 818: I think we should add a message here if it works. Like it does in the `login`

dds_cli/__main__.py
```diff
@@ -742,40 +768,62 @@ def create(
...
     try:
-        with dds_cli.session.Session(username=username, check=check) as session:
-            if check:
-                session.check()
-            else:
-                # Session renewed in the init method.
-                LOG.info("[green] :white_check_mark: Session renewed![/green]")
+        with dds_cli.auth.Auth(username=username) as authenticator:
+            # Authentication token renewed in the init method.
+            LOG.info("[green] :white_check_mark: Authentication token renewed![/green]")
```
inaod568, line 800: We can skip the context manager here and just do `dds_cli.auth.Auth(username=username)` since the `authenticator` isn't used. Or alternatively `with dds_cli.auth.Auth(username=username):` ( without the "as authenticator" part)

inaod568 (commented):
Nice! Looks and works great, just a few comments and suggestions. There's currently an error when attempting to upload but that has nothing to do with this and should be fixed when my s3-info PR is merged.

dds_cli/user.py
alneberg: Yes, it's just that I wanted to expose it to be used in the auth commands. So in the `auth.py` it would be `token_file.token_file.exists()` with your suggestion instead of `token_file.file_exists()` now. I would be fine with both though.

dds_cli/__main__.py
alneberg: Done!

dds_cli/__main__.py
alneberg: I went for the second option.

inaod568 (approved):
Looks great and now it all works.

PR #199, opened 2021-11-25
Cli tests for add user and ls

alneberg:
Running tests independently of the API where all http requests are mocked and return values are explicitly set in the test files. This will of course need to be updated when we change the API. The isolation of the cli in the tests can also be seen as a benefit since it's much easier to debug when something goes wrong.

The mocking is done in pytest fixtures and the actual commands are triggered with the utilities for testing supplied with click. To keep the mocking flexible where the return values can be modified between tests, the fixtures return a function that in turn returns a mock when called. When calling this method, the mock can be customised.

Tests for other commands will come along the way.


alneberg:
Test runs in github actions! 🎉 
Screenshot 2021-11-25 at 10 30 37

inaod568 (approved):

PR #200, opened 2021-11-26
Issue703: Introduce dds user command group

MatthiasZepper:
This PR comprises the necessary changes to the CLI to address [issue 703](https://github.com/ScilifelabDataCentre/dds_web/issues/703) from the web: 

- Newly introduced the feature to delete user accounts in the CLI.
- To delete the own account, one would need to enter `dds user delete --self`
-  To delete any other account (given sufficient administrative privileges), one would need to specify `dds user delete EMAIL`
- The former `dds add-user`  command is now `dds user add` and its former option `--email` is now its argument.
- Created an `utils.email_validator` function that will employ a simple Regex to check if a valid email address was entered before dispatching the API call since a failed marshmallow validation doesn't seem to be properly handled in the CLI so far.

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -1044,7 +1044,7 @@ def add(click_ctx, username, email, role, project):
 
 
 ## Delete users from the system
-@user.command(no_args_is_help=True)
+@user.command(name="delete", no_args_is_help=True)
 @click.argument(
     "email",
     nargs=1,
@@ -1059,7 +1059,7 @@ def add(click_ctx, username, email, role, project):
     help="Decommission your own user account",
 )
 @click.pass_obj
-def delete(click_ctx, email, self):
+def delete_users(click_ctx, email, self):
     """
     Delete user accounts from the Data Delivery System.
 
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -1014,8 +1014,15 @@ def user(click_ctx):
     help="Existing Project you want the user to be associated to.",
 )
 @click.pass_obj
-def invite(click_ctx, username, email, role, project):
-    """Add a user to DDS, sending an invitation email to that person."""
+def add(click_ctx, username, email, role, project):
+    """
+    Add a user to the DDS system or hosted projects.
+
+    Specify an user's email and role to associate it with projects.
+    If the user doesn't exist in the system yet, an invitation email
+    will be sent automatically to that person.
+
+    """
     try:
         with dds_cli.account_manager.AccountManager(
             username=username, no_prompt=click_ctx.get("NO_PROMPT", False)
@@ -1073,7 +1080,6 @@ def delete(click_ctx, email, self):
     try:
         with dds_cli.account_manager.AccountManager(
             username=None,
-            authenticate=True,
             method="delete",
             no_prompt=click_ctx.get("NO_PROMPT", False),
         ) as manager:
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -982,15 +982,19 @@ def user(click_ctx):
 
 ## Invite new users to the system
 @user.command()
+@click.argument(
+    "email",
+    nargs=1,
+    type=str,
+    required=True,
+    callback=dds_cli.utils.email_validator,
+)
 @click.option(
     "--username",
     "-u",
     required=False,
     type=str,
     help="Your Data Delivery System username.",
-)
-@click.option(
-    "--email", "-e", required=True, type=str, help="Email of the user you would like to invite."
 )
 @click.option(
     "--role",
@@ -1033,8 +1037,14 @@ def invite(click_ctx, username, email, role, project):
 
 
 ## Delete users from the system
-@user.command()
-@click.argument("accounts", nargs=-1)
+@user.command(no_args_is_help=True)
+@click.argument(
+    "email",
+    nargs=1,
+    type=str,
+    required=False,
+    callback=dds_cli.utils.email_validator,
+)
 @click.option(
     "--self",
     required=False,
@@ -1042,7 +1052,7 @@ def invite(click_ctx, username, email, role, project):
     help="Decommission your own user account",
 )
 @click.pass_obj
-def delete(click_ctx, accounts, self):
+def delete(click_ctx, email, self):
     """
     Delete user accounts from the Data Delivery System.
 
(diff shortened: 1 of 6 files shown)


dds_cli/utils.py
```diff
@@ -99,3 +101,16 @@ def format_api_response(response, key, magnitude=None, iec_standard=False):
...
+    try:
+        while value is not None:
+            if not re.match(r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)", value):
+                raise ValueError(value)
+            else:
+                return value
+    except ValueError as err:
+        click.echo("Incorrect email address given: {}".format(err))
+        value = click.prompt("Please try again")
+        return email_validator(ctx, param, value)
```
inaod568, line 116: Are we sure there's not a package for this that we can use?

dds_cli/__main__.py
```diff
@@ -1002,3 +946,128 @@ def abort(click_ctx, username, project):
...
+    If you have sufficient administrative privileges, you may also delete the accounts of other users.
+    Specify the e-mail address as argument to the main command to initiate the removal process.
+    """
+    if email:
+        # Click.confirm unfortunately doesn't support formatted or colored strings, so use Log to warn.
+        LOG.info(f"[red]Delete Data Delivery System user account associated with {email}?[/red]")
+    else:
+        LOG.info(f"[red]Are you sure? Deleted accounts can't be restored![/red]")
+
+    click.confirm("Proceed with account deletion?", abort=True)
```
inaod568, line 1056: https://rich.readthedocs.io/en/stable/prompt.html

dds_cli/__main__.py
```diff
@@ -1002,3 +946,128 @@ def abort(click_ctx, username, project):
...
+
+    try:
+        with dds_cli.account_manager.AccountManager(
+            username=None,
+            method="delete",
+            no_prompt=click_ctx.get("NO_PROMPT", False),
+        ) as manager:
+            manager.delete_user(email=email, ownaccount=self)
```
inaod568, line 1064:
```
if self:
    manager.delete_own_account()
else: 
    manager.delete_user(email=email)
```

dds_cli/__main__.py
```diff
@@ -1002,3 +946,128 @@ def abort(click_ctx, username, project):
...
+    """
+    Delete user accounts from the Data Delivery System.
+
+    To request the removal of your own account, use the `--self` flag without any arguments.
+    An e-mail will be sent to you asking to confirm the deletion.
+
+    If you have sufficient administrative privileges, you may also delete the accounts of other users.
+    Specify the e-mail address as argument to the main command to initiate the removal process.
+    """
+    if email:
```
inaod568, line 1050: Add a check to make sure that either `email` or `self` is used

inaod568 (changes requested):

dds_cli/utils.py
MatthiasZepper: Who is we? And is there any specific reason why this solution is wrong or unacceptable?

dds_cli/__main__.py
```diff
@@ -1002,3 +946,128 @@ def abort(click_ctx, username, project):
...
+
+## Delete users from the system
+@user.command(no_args_is_help=True)
+@click.argument(
+    "email",
+    nargs=1,
+    type=str,
+    required=False,
+    callback=dds_cli.utils.email_validator,
+)
```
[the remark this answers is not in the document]
MatthiasZepper: The purpose of which would be? 

1.  Like in `dds ls -u username` the username of the person logged in? 
2.  Or the username to be deleted from the system?

I considered case _1._ and opted against it because I believe that would be highly confusing with regard to user experience, because the final command would look like `dds user delete -u [my_username_i_like_to_keep] [email_of_the_user_to_delete]`.

dds_cli/__main__.py
MatthiasZepper: Sure. I just have had a look how prompts are implemented in other functions, but I can use rich instead.

dds_cli/__main__.py
MatthiasZepper: If neither is specified, that would trigger the help message, since the command reads `@user.command(name="delete", no_args_is_help=True)`.

dds_cli/utils.py
inaod568: It's not wrong or unacceptable. It's nicely written. But the general idea is to "not reinvent the wheel" so if there's another package (maybe it's possible to use the `marshmallow.validator.Email` but I'm not sure if it's usable here) or function in a already used package that does this exact thing (checks if an email has the correct format) then we should use that and not write our own function.

inaod568: Yes. The solution is sufficient and no it's not rocket science. All I wanted to know was if there was another shorter way of doing this, and since there apparently isn't then using this is fine.

dds_cli/account_manager.py
```diff
@@ -74,3 +79,62 @@ def add_user(self, email, role, project):
...
+                raise dds_cli.exceptions.DDSCLIException(message)
+        else:
+            LOG.info(message)
+
+    def delete_own_account(self):
+        """Delete users from the system"""
+        # Perform request to API
+
+        try:
+            response = requests.post(
```
inaod568, line 118: Not a dead breaker but I think we should change `post` to `delete` here (and of course in the api)

inaod568: We can do this in another PR though

dds_cli/utils.py
```diff
@@ -4,6 +4,12 @@
 import rich.console
 import simplejson
 
+import click
+import rich.console
+import numbers
+import rich.console
+import numbers
```
inaod568, line 11: duplicates

dds_cli/utils.py
```diff
@@ -4,6 +4,12 @@
 import rich.console
 import simplejson
 
+import click
```
inaod568, line 7: unused

inaod568 (approved):
Works well and looks good. A few comments and things that should be removed or changed, but I don't feel they are deal breakers and I'm happy to fix these in other PRs instead of waiting for you to do this (since you're on holiday). So I'm approving and merging, unless anyone (@alneberg @aanil )  has any objections?

dds_cli/account_manager.py
MatthiasZepper: Fixed. Will open the nano-PR to web soon!

PR #203, opened 2021-12-07
Use requests for download

inaod568:
* Changed the download process to use presigned urls with the python requests package instead of boto3
* Removed the bucket check and generation 
* Due do changes in the file info there needs to be changes in the summary tables, removed this for now


alneberg (approved):
Very nice, some nice cleanup as well! Seems to be working fine!

PR #206, opened 2021-12-09
Removed all umask

inaod568:
Removing all os.umask() occurrences since it doesn't seems to do anything. 

Please test this yourselves though and let me know if something regarding reading or writing breaks for you.


alneberg (approved):
Very welcome change!

PR #207, opened 2021-12-10
Logging vs progressbars

alneberg:
1. Switch back rich logging level to info (unless verbose)
2. Make progress bars log to stderr instead of stdout
3. Move some log statements from info to debug if they’re normally not needed for the end user (This is of course a subjective choice)

This will re-enable the messages printed to users for `dds auth info` and `dds ls` which are otherwise muted if the level is set to warning.


inaod568 (approved):
Nice! Looks great 👍🏻

PR #210, opened 2021-12-20
Add command project, subcommands grant&revoke and restructure code a bit

aanil:
* Created `grant` and `revoke` commands to grant and revoke project access to users. `grant` is `add-user --project` split from `add-user` and renamed.
* Created `project` command and grouped `create`, `grant`, `revoke` and `status` under it.
* Changed the option `--username` to use a common option


dds_cli/__main__.py
```diff
@@ -649,17 +644,90 @@ def get(
...
+
+
+@auth.command()
 @click.option(
     "--username",
     "-u",
     required=False,
     type=str,
-    help="Your Data Delivery System username.",
+    help="Your Data Delivery System username. Required unless the `--check` flag is used.",
```
inaod568, line 662: What does the `--check` flag do? Is this something new you're implementing?

aanil: No, it was already there. I just reordered the auth commands, didn't add anything.

inaod568: Oh ok. Then git is doing something strange since it's saying that it's added in this PR. But anyway, then you can remove this, we don't know what this is. 👍🏻

inaod568 (approved):

PR #213, opened 2022-01-03
Group commands

inaod568:


dds_cli/__main__.py
```diff
@@ -993,37 +757,68 @@ def archive(click_ctx, username, project):
...
 
 
-@status.command()
-@common_options
-@common_options_project
+# ACCESS ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ACCESS #
+@project_group_command.group(name="access")
+@click.pass_obj
+def project_access(_):
+    """Manage project statuses."""
```
aanil, line 764: Maybe a typo? Should be managing access instead of status here.

dds_cli/__main__.py
```diff
@@ -372,356 +264,243 @@ def ls(click_ctx, project, folder, projects, size, username, usage, sort, tree,
...
+####################################################################################################
+####################################################################################################
+## PROJECT ############################################################################## PROJECT ##
+####################################################################################################
+####################################################################################################
 
 
-@auth.command()
-@common_options
+@dds_main.group(name="project", invoke_without_command=True)
```
aanil, line 475: Maybe `no_args_is_help=True` could be added here too.

dds_cli/__main__.py
```diff
@@ -744,24 +523,25 @@ def common_options_project(f):
...
+    help="Email of a user to be added to the project as Researcher",
 )
+# Flags
 @click.option(
     "--owner",
+    "owner",
     required=False,
     multiple=True,
-    help="Email of a user to be added to the project as Project Owner",
+    help="Email or user to be added to the project as Project Owner",
```
aanil, line 537: Maybe a typo? `Email of user ... `

dds_cli/__main__.py
```diff
@@ -744,24 +523,25 @@ def common_options_project(f):
     help="The name of the Principal Investigator",
 )
 @click.option(
```
aanil, line 525: click doesn't really specify in the cli that this option can be used multiple times, so should we put that detail in the description?

dds_cli/__main__.py
```diff
@@ -1033,134 +828,323 @@ def abort(click_ctx, username, project):
...
-###################################################################################
-## Invite new users to the system
-@user.command()
-@click.argument(
-    "email",
-    nargs=1,
-    type=str,
-    required=True,
+def data_group_command(_):
+    """Group command: dds project. Manage projects."""
```
aanil, line 841: `dds project. Manage projects` maybe a type since the command is `dds data `here?

dds_cli/__main__.py
```diff
@@ -1033,134 +828,323 @@ def abort(click_ctx, username, project):
...
-###################################################################################
-@dds_main.group(no_args_is_help=True)
+####################################################################################################
+####################################################################################################
+## DATA #################################################################################### DATA ##
+####################################################################################################
+####################################################################################################
+
+
+@dds_main.group(name="data", invoke_without_command=True)
```
aanil, line 838: Could have `no_args_is_help=True` here too.

dds_cli/__main__.py
```diff
@@ -372,356 +264,243 @@ def ls(click_ctx, project, folder, projects, size, username, usage, sort, tree,
...
+# -- dds project ls -- #
+@project_group_command.command(name="ls")
+# Options
+@username_option()
+@sort_projects_option()
+# Flags
+@usage_flag(help_message="Show the usage for available projects, in GBHours and cost. ")
+@json_flag(help_message="Output project list as json.")  # users, json, tree
+@click.pass_context
+def list_projects(ctx, username, json, sort, usage):
```
aanil, line 495: Should this list all projects or just a single project since its a subcommand of `dds project`. `dds data ls` takes a single project.

dds_cli/__main__.py
```diff
@@ -1033,134 +828,323 @@ def abort(click_ctx, username, project):
...
+@project_option(required=True)
+@folder_option(help_message="List contents in this project folder.")
+# Flags
+@json_flag(help_message="Output in JSON format")
+@size_flag(help_message="Show size of project contents.")
+@tree_flag(help_message="Display the entire project(s) directory tree")
+@users_flag(help_message="Display users associated with a project(Requires a project id)")
+@click.pass_context
+def list_data(ctx, username, project, folder, json, size, tree, users):
+    """List project contents. Same as dds ls [PROJECT ID]."""
```
MatthiasZepper, line 1070: I don't think dds ls accepts the project ID as an argument? At least the decorator is missing if you intend to have this functionality.

dds_cli/__main__.py
```diff
@@ -372,356 +264,243 @@ def ls(click_ctx, project, folder, projects, size, username, usage, sort, tree,
...
-    return functools.reduce(lambda x, opt: opt(x), options, f)
+# -- dds project ls -- #
+@project_group_command.command(name="ls")
+# Options
+@username_option()
+@sort_projects_option()
+# Flags
+@usage_flag(help_message="Show the usage for available projects, in GBHours and cost. ")
+@json_flag(help_message="Output project list as json.")  # users, json, tree
+@click.pass_context
```
MatthiasZepper, line 494: I think, this is done on purpose because it is the sole non-group command on the top-level, but just to double-check: You deliberately used `click.pass_context` here instead of `click.pass_obj`?

MatthiasZepper (approved):
Well done! I would rephrase some of the help messages slightly, but I guess this is precision work left for later.

dds_cli/__main__.py
inaod568: Fixed!

dds_cli/__main__.py
inaod568: Added that. But removed the `invoke_without_command`. Unnecessary.

dds_cli/__main__.py
inaod568: Correct. `click.pass_obj` passes a dict, and `click.pass_context` passes an object that can then perform the `invoke` of another command.

dds_cli/__main__.py
inaod568: In the `dds-development` chat we agreed on that `dds project ls` should list projects and therefore call `dds ls`. 
```
`dds project`
    `dds project ls` --> list projects — CALLS `dds ls` command
    
`dds data`
    `dds data ls <options>` --> list project contents -- CALLS `dds ls [PROJECT ID]` command
```

dds_cli/__main__.py
inaod568: Fixed!

dds_cli/__main__.py
inaod568: Fixed!

dds_cli/__main__.py
inaod568: I'm not sure I follow, just gonna reply a bit here, hope this clears it up:

This is `dds data ls` which takes the project id as an option. So the call should be `dds data ls --project`. This will list the project contents by calling the `dds ls` command which takes the project option too. Before it was an argument but in the `dds-development` channel we said that we should have the project ID as _either_ an positional argument or an option, and since it doesn't really make sense to have it as an argument in most cases, or at least just adds a bit to the confusion, I decided to add the `project_option` to `dds ls` instead.

MatthiasZepper: I was just referring to the help text that I found to be in disagreement with the way the command worked and wanted to make sure that the command is not intended to accept an argument. 

Yesterday I went though the various subcommands and tried to spot if I noticed any missing options or help texts that didn't fit the functionality. Here, I would just have expected _Same as dds ls -p [PROJECT ID]_ - so nothing serious!

dds_cli/__main__.py
inaod568: Fixed!

inaod568:
> Well done! I would rephrase some of the help messages slightly, but I guess this is precision work left for later.

Thanks! Yes, we can work on those at a later time 👍🏻

aanil (approved):
👍 Looks good!

PR #214, opened 2022-01-05
Add dds user info command

aanil:
Displays a logged in user's name, username, role and associated emails

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -476,21 +476,18 @@ def get_info_user(click_ctx, username):
     Display info about the user logged in from the Data Delivery System.
 
     """
-    if click_ctx.get("NO_PROMPT", False):
-        pass
-    else:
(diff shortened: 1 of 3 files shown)


dds_cli/__main__.py
```diff
@@ -465,6 +465,34 @@ def delete_user(click_ctx, email, username, self):
...
+# Flags
+@click.pass_obj
+def get_info_user(click_ctx, username):
+    """
+    Display info about the user logged in from the Data Delivery System.
+
+    """
+    if click_ctx.get("NO_PROMPT", False):
+        pass
+    else:
```
inaod568, line 481: This should be removed, right? This means that we only get the user info if the no-prompt flag is not used, and then the no_prompt variable is used in like 484 anyway.

inaod568 (approved):
Looks good!

PR #217, opened 2022-01-12
Add timestamp localization by using local timezone info

aanil:
Converts the timestamps in UTC received from dds_web to timestamps in the local timezone on the system where the CLI is installed. Should work on UNIX-like and Windows systems, although not tested for Windows.

### DIFF
--- a/dds_cli/project_status.py
+++ b/dds_cli/project_status.py
@@ -4,6 +4,9 @@ import logging
 # Installed
 import requests
 import simplejson
+import pytz
+import tzlocal
+import datetime
 
 # Own modules
 from dds_cli import base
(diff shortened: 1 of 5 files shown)


dds_cli/project_status.py
```diff
@@ -66,10 +69,28 @@ def get_status(self, show_history):
...
         else:
-            LOG.info(f"Current status of {self.project}: {resp_json.get('current_status')}")
+            current_status = resp_json.get("current_status")
+            current_deadline = resp_json.get("current_deadline")
+            status_out = f"Current status of {self.project}: {current_status}"
+            deadline_out = ""
+            if current_deadline:
+                date = pytz.timezone("UTC").localize(
+                    datetime.datetime.strptime(current_deadline, "%a, %d %b %Y %H:%M:%S GMT")
+                )
```
inaod568, line 79: Since this generated an error due to GMT before, maybe we should add a try/except block to catch that error and in that case just print the time or something?

inaod568: Just so we avoid a long exception that the user doesn't understand

dds_cli/project_status.py
```diff
@@ -66,10 +69,28 @@ def get_status(self, show_history):
...
+                    "%a, %d %b %Y %H:%M:%S %Z"
+                )
+                deadline_out = f" with deadline {current_deadline}"
+            LOG.info(f"{status_out}{deadline_out}")
             if show_history:
                 history = "Status history \n"
                 for row in resp_json.get("history"):
+                    date = pytz.timezone("UTC").localize(
+                        datetime.datetime.strptime(row[1], "%a, %d %b %Y %H:%M:%S GMT")
+                    )
```
inaod568, line 90: Same here and the other places we do this in

PR #218, opened 2022-01-14
Add sensitive checks

inaod568:


inaod568:
Going to start from scratch with this. Probably stealing a bit of this code. Not a prio atm though.

PR #222, opened 2022-01-21
Enable use of HOTP for authenticating

alneberg:


dds_cli/user.py
```diff
@@ -99,12 +99,48 @@ def __authenticate_user(self):
...
+                    message="Exited due to no one-time authentication code entered."
+                )
+
+            if not token_entered.isdigit():
+                LOG.info("Please enter a valid one-time code. It should consist of only digits.")
+                continue
+            if len(token_entered) != 8:
+                LOG.info("Please enter a valid one-time code. It should consist of 8 digits.")
+                continue
+            one_time_password = token_entered
```
inaod568, line 136: I think this works really well from a users perspective, but I'm wondering if it's not possible to return the token instead of doing basic auth both times?

worukan: I agree that the user experience of the login with mfa is carefully implemented. However, I think that a user should not be able to request 2fa, instead it is something enforced by the server. Moreover, I agree with Ina that providing user crendentials should follow with a token returned to confirm that the first factor is successfully authenticated, but we need an additional factor of authentication to follow up. For the second factor, use the token from the first factor to create the final token to be used for normal operation of the services for the given user.

dds_cli/user.py
```diff
@@ -112,13 +111,64 @@ def __authenticate_user(self):
...
         if not response.ok:
-            message = response_json.get("message", "Unexpected error!")
             if response.status_code == 401:
-                raise exceptions.AuthenticationError(message=message)
+                raise exceptions.AuthenticationError(
+                    "Authentication failed, incorrect username and/or password."
+                )
+            else:
+                raise dds_cli.exceptions.ApiResponseError(
+                    message=f"API returned an error: {response_json['message']}"
```
inaod568, line 121: I'd change to `response_json.get('message', "Unknown Error!")`

dds_cli/user.py
```diff
@@ -112,13 +111,64 @@ def __authenticate_user(self):
...
+
+        # Token received from API needs to be completed with a mfa timestamp
+        partial_auth_token = response_json.get("token")
+
+        # Verify 2fa email token
+        LOG.info(
+            "Please enter the one-time authentication code sent to your email address (leave empty to exit):"
+        )
+        done = None
+        while done is None:
```
inaod568, line 132:
```
done = False
while not done:
```
instead? Since you're setting `done = True` further down this makes more sense I think.

dds_cli/user.py
alneberg: Agreed, fixed!

inaod568 (approved):

PR #227, opened 2022-02-01
Token lifetime can now be displayed from the tokens header

MatthiasZepper:
Previously, the CLI relied on the creation/modification date of the `~./dds_cli_token` file and a locally contained configuration value `TOKEN_MAX_AGE` to estimate the validity of the token. 

Now, the CLI will only fall back to `TOKEN_MAX_AGE` and the modification date if no expiration time is specified in the token's header.

Additionally, the `TOKEN_WARNING_AGE` has been changed from an absolute value to a fraction of the lifetime respectively `TOKEN_MAX_AGE`.

The custom `get_token_header_contents()` function was written because the [`get_unverified_header()`](https://pyjwt.readthedocs.io/en/stable/usage.html?highlight=get_unverified_header#reading-headers-without-validation) from the PyJWT package fails with`jwt.exceptions.DecodeError: Invalid payload padding` for our encrypted tokens. The reason is that a standard JWT token contains three parts separated by a dot: `header.payload.signature`, whereas our tokens follow a non-standard `header.empty_payload.secondary_header.encrypted_payload.encrytped_signature` pattern.

### DIFF
--- a/dds_cli/user.py
+++ b/dds_cli/user.py
@@ -265,23 +265,17 @@ class TokenFile:
         what to do.
 
         Returns True if the token has expired, False otherwise.
-
-        ** Notice, this does not actually know the expiration time of the token, it's possible that the
-        expiration time has changed upstream in which case the variables need to be updated.
-
-        if check is True, token age will be reported to stdout
         """
-        age, lifetime, _ = self.__token_dates(token_metadata=token_metadata)
-        LOG.debug(f"Token file age: {age}")
-        if age > lifetime:
-            LOG.debug(
-                "Token file is too old so token has likely expired. Now deleting it and fetching new token."
-            )
+        token_dates = self.__token_dates(token_metadata=token_metadata)
+
+        LOG.debug(f"Token file age: {readable_timedelta(token_dates['age'])}")
+        if token_dates["lft"].total_seconds() >= 0:
+            LOG.debug("Token has expired. Now deleting it and fetching new token.")
             self.delete_token()
             return True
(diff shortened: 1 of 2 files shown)


inaod568:
@MatthiasZepper I set this as ready to review so that I can have a look and test when I have time. Will do the same for the web. Will let you know if there's anything I need from you regarding this, otherwise keep working on the other tasks. (Even if there is a conflict in the web.)

dds_cli/user.py
```diff
@@ -145,22 +147,29 @@ def read_token(self):
...
         with self.token_file.open() as file:
             token = file.read()
             if not token:
                 raise exceptions.TokenNotFoundError(message="Token file is empty.")
 
+            # Use lifetime from token header if given, else read default from config
+        try:
+            lft = jwt.get_unverified_header(token).get("lft")
+            lifetime = isodate.parse_duration(lft) if lft else dds_cli.TOKEN_MAX_AGE
+        except:
```
alneberg, line 164: It's a general recommendation to avoid `except` clauses like this. The recommendation is to specify the exception you except to be raised instead. The major benefit I can see is that it's much easier to refactor it in the future.

MatthiasZepper: Thanks for the helpful advice! Then I guess I will add the base exception [`jwt.exceptions.InvalidTokenError`](https://pyjwt.readthedocs.io/en/latest/api.html#exceptions) or should I be super specific and really add the relevant subclasses?

PR #233, opened 2022-02-11
Click help: max 100 chrs wide

ewels:
I changed the default maximum width for `rich-click` to be the full width of the terminal.

This PR customises the package to limit the width to 100 characters (or the width of the terminal if less than that).

Personally I think it looks better for `dds` as the majority (all?) of the help messages fit within this width.

The end result is the same as when I originally added the package, before I changed the default.

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -59,6 +59,7 @@ from dds_cli.options import (
 LOG = logging.getLogger()
 
 # Monkey patch click to use rich's logging
+rich_click.core.MAX_WIDTH = 100
 click.Group.format_help = rich_click.rich_format_help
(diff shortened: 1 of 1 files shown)


inaod568 (approved):

inaod568:
I tried it out and didn't actually notice a difference, so as you say most things should fit within 100 chars. 👍🏻

ewels:
Ah right, I hadn't pushed the release yet which is why you didn't see any change in behaviour 😅

inaod568:
@ewels 😆 Okey. Have you done that now? Let me know and I'll try it out.

ewels:
Yup, v0.3.0 went out last night! Still fairly heavy development so it's very likely that I'll need to do another PR soon 👀

PR #237, opened 2022-02-16
Change log and printout info after delivery

inaod568:


dds_cli/file_handler.py
```diff
@@ -56,12 +56,19 @@ def __init__(self, user_input, local_destination, project=None):
...
+                "status": FileHandler.make_json_serializable(non_json=status),
+            }
+        }
         try:
-            with file.open(mode="a") as errfile:
+            with log_file.open(mode="a") as errfile:
                 json_output = json.dumps(
-                    info,
+                    failed_to_save,
                     indent=None,
```
aanil, line 72: Maybe provide an indent value so the file is a bit easier to read?

inaod568: Done!

aanil (approved):
👍 Looks good!

PR #244, opened 2022-02-18
Show logged in user in header

aanil:
Before submitting a pr:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -32,6 +32,7 @@ import dds_cli.directory
 import dds_cli.project_creator
 import dds_cli.auth
 import dds_cli.project_status
+import dds_cli.user
 import dds_cli.utils
 from dds_cli.options import (
     email_arg,
@@ -74,6 +75,8 @@ click.Command.format_help = rich_click.rich_format_help
 #                                                                                                  #
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # ##
 
+# Get token metadata
+_, metadata = dds_cli.user.TokenFile().read_token()
 
 # Print header to STDERR
 dds_cli.utils.stderr_console.print(
@@ -82,7 +85,8 @@ dds_cli.utils.stderr_console.print(
     "\n[green](  ) ) (  (  )[/]   [bold]SciLifeLab Data Delivery System",
     "\n[green] ︶  (  ) ) ([/]    [blue][link={0}]{0}[/link]".format(dds_cli.__url__),
     f"\n[green]      ︶ (  )[/]    [dim]Version {dds_cli.__version__}",
-    "\n[green]          ︶\n",
+    "\n[green]          ︶",
+    f"\n[green]Logged in:[/] [red]{metadata.get('csg') if metadata else 'None'}",
     highlight=False,
 )
 
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -83,7 +83,7 @@ dds_cli.utils.stderr_console.print(
     "\n[green] ︶  (  ) ) ([/]    [blue][link={0}]{0}[/link]".format(dds_cli.__url__),
     f"\n[green]      ︶ (  )[/]    [dim]Version {dds_cli.__version__}",
     "\n[green]          ︶",
-    f"\n[green]Current user:[/] [red]{username}",
+    f"\n[green]Current user:[/] [red]{username}" if username else "",
     highlight=False,
 )
 
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -73,7 +73,7 @@ click.rich_click.MAX_WIDTH = 100
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # ##
 
 # Get token metadata
-_, metadata = dds_cli.user.TokenFile().read_token()
+username = dds_cli.user.User.get_user_name_if_logged_in()
 
 # Print header to STDERR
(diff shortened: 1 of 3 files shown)


inaod568:
@worukan Is working on changing the token after at the moment and displaying information from the token in the CLI, so I think we'll wait a bit with this

aanil:
@worukan I'm not sure I understand. The user won't log in every time they use the CLI. They will login once and the token will be saved on their system and they'll use the token to run commands until the token expires?

inaod568:
I don't think displaying the username when during a command is a `must have`, more like a `could have`. If the user wants to know what username is authenticated in the current session, they can run the `dds auth info` command as @worukan said. So I think we can skip that for now and if we feel like this is something we want to work on in a few weeks or so, then we can do that.

worukan:
Perhaps the cli can remember the username or store it with the token in the token file?

MatthiasZepper:
> If the user wants to know what username is authenticated in the current session, they can run the `dds auth info` command as @worukan said. 

Not anymore. This functionality is scrapped entirely at the moment. 

Also we do not store anything locally except the `.dds_cli_token` file. So unless we update the format to let's say a JSON file that can locally cache more values than just the token, I don't see how to store the username for the lifetime of the token if not within the token itself.

inaod568:
@MatthiasZepper Since we're not able to test at the moment, I can't have a detailed discussion about this. But, the username is always required by the token, otherwise we would never know the `current_user()` in the web. So if this functionality is completely scrapped now, then we can add that at a later time, by for example calling the API, authenticating with the token as usual and just returning the username. This should be something fairly easy to implement too and we can even do that before the testing starts probably. So I think we can continue with the changes that Volkan is making and we'll see what possible additions and future improvements we can do later.

aanil:
Changed the solution to get username from API.

MatthiasZepper:
> Changed the solution to get username from API.

Ah nice. I completely forgot that there already is an endpoint to display the user information!

dds_cli/__main__.py
```diff
@@ -79,7 +82,8 @@
     "\n[green](  ) ) (  (  )[/]   [bold]SciLifeLab Data Delivery System",
     "\n[green] ︶  (  ) ) ([/]    [blue][link={0}]{0}[/link]".format(dds_cli.__url__),
     f"\n[green]      ︶ (  )[/]    [dim]Version {dds_cli.__version__}",
-    "\n[green]          ︶\n",
+    "\n[green]          ︶",
+    f"\n[green]Logged in:[/] [red]{username}",
```
inaod568, line 86: Maybe we could change this to something like 
* User: 
* Authenticated user:
* Running command as: 
* Using dds-cli as
Not sure but "Logged in" doesn't feel right to me

alneberg:
@worukan what's wrong with putting the username in the readable token header? The solution as it stands now will have implications for authentication where an HOTP authentication will be initiated if the token has expired. And potentially one would hit the request limit if one does 10 `--help` commands with an expired token. I guess one could save the username in the tokenfile and change the format of the tokenfile but then we rely on unsigned information created by the CLI instead of the signed header of the token.

worukan:
> @worukan what's wrong with putting the username in the readable token header? The solution as it stands now will have implications for authentication where an HOTP authentication will be initiated if the token has expired. And potentially one would hit the request limit if one does 10 `--help` commands with an expired token. I guess one could save the username in the tokenfile and change the format of the tokenfile but then we rely on unsigned information created by the CLI instead of the signed header of the token.

Because it is the wrong place. That token header is mainly for being able to decrypt the token correctly. Just because we can put information in there does not justify doing so. The encrypted token is a special token. It should not be seen as an all-purpose token.

If you like I have a new suggestion @alneberg @MatthiasZepper @aanil @inaod568 
We used to have a token endpoint which provided a signed token with sub and exp in it. While the endpoint is not available now, the functionality is still there which is `jwt_token` function. We could bring the endpoint back easily.

1. Request `jwt_token` from the cli.
2. Request `encrypted_jwt_token` from the cli.
3. Save `jwt_token` and use the expiration time and the username from that.
4. Save `encrypted_jwt_token`. If it simplifies, they can be saved to separate files.
5. Use the `encrypted_jwt_token` for normal operation.

In this way the expiration time of the encrypted token will come slightly after the signed token, but that is a negligible difference.

alneberg:
Right, I see. Although I also realised that just skipping the API call for user info in case the token is expired would solve most of my issues as well.

alneberg (approved):
Looks good! I might have chosen a different color than red for the username, but that can be changed later.

dds_cli/__main__.py
```diff
@@ -79,7 +82,8 @@
     "\n[green](  ) ) (  (  )[/]   [bold]SciLifeLab Data Delivery System",
     "\n[green] ︶  (  ) ) ([/]    [blue][link={0}]{0}[/link]".format(dds_cli.__url__),
     f"\n[green]      ︶ (  )[/]    [dim]Version {dds_cli.__version__}",
-    "\n[green]          ︶\n",
+    "\n[green]          ︶",
+    f"\n[green]Current user:[/] [red]{username}",
```
inaod568, line 86: Since row 76 might return `None` if something fails in the API call to get the username, I'm thinking we should have the row here as 
```
f"\n[green]Current user:[/] [red]{username}" if username else None
```

inaod568: Otherwise it's going to show up as `Current user: None`

inaod568 (changes requested):
Looks

dds_cli/__main__.py

inaod568 (approved):
Looks good!

PR #246, opened 2022-02-19
change to non-sensitive

inaod568:
Before submitting a pr:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -651,7 +651,7 @@ def list_projects(ctx, username, json, sort, usage):
     "--non-sensitive",
     required=False,
     is_flag=True,
-    help="Indicate if the Project includes sensitive data.",
+    help="Indicate whether the project contains only non-sensitive data",
 )
 @click.pass_obj
 def create(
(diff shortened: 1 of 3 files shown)


dds_cli/__main__.py
```diff
@@ -648,7 +648,7 @@ def list_projects(ctx, username, json, sort, usage):
     + dds_cli.utils.multiple_help_text(item="project owner"),
 )
 @click.option(
-    "--is_sensitive",
+    "--non-sensitive",
     required=False,
     is_flag=True,
     help="Indicate if the Project includes sensitive data.",
```
alneberg, line 654: I would probably rephrase the help message slightly as well.

talavis: "Indicate whether the project contains only non-sensitive data"

alneberg (approved):
Great! I would think @talavis suggestion for the help message is an improvement, but no deal breaker.

dds_cli/__main__.py
inaod568: Will fix 👍🏻

inaod568: Done

PR #251, opened 2022-02-21
Token expiration

worukan:
Before submitting a pr:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

### DIFF
--- a/dds_cli/user.py
+++ b/dds_cli/user.py
@@ -21,7 +21,7 @@ import rich
 # Own modules
 import dds_cli
 from dds_cli import exceptions
-from dds_cli.utils import get_token_header_contents, readable_timedelta
+from dds_cli.utils import get_token_expiration_time, readable_timedelta
 
 ###############################################################################
 # START LOGGING CONFIG ################################# START LOGGING CONFIG #
@@ -65,7 +65,7 @@ class User:
             # Get token from file
             try:
                 LOG.debug(f"Checking if token file exists for user {self.username}")
-                self.token, _ = token_file.read_token()
+                self.token = token_file.read_token()
             except dds_cli.exceptions.TokenNotFoundError:
                 self.token = None
 
@@ -198,7 +198,7 @@ class TokenFile:
 
         if not self.file_exists():
             LOG.debug(f"Token file {self.token_file} does not exist.")
-            return None, None
+            return None
 
         self.check_token_file_permissions()
 
@@ -208,18 +208,12 @@ class TokenFile:
             if not token:
                 raise exceptions.TokenNotFoundError(message="Token file is empty.")
 
-            # Use lifetime from token header if given, else read default from config
-        try:
-            token_metadata = get_token_header_contents(token)
-        except exceptions.TokenNotFoundError:
-            token_metadata = None
-
-        if self.token_expired(token_metadata=token_metadata):
+        if self.token_expired(token=token):
             LOG.debug("No token retrieved from file, will fetch new token from API")
-            return None, None
+            return None
 
         LOG.debug("Token retrieved from file.")
-        return token, token_metadata
+        return token
 
     def file_exists(self):
         """Returns True if the token file exists."""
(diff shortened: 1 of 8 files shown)


dds_cli/user.py
```diff
@@ -312,40 +309,30 @@ def token_report(self, token_metadata=None):
...
+            LOG.info(f"[{markup_color}]Token expired: {expiration_time}[/{markup_color}]")
+        else:
+            lifetime = age - dds_cli.TOKEN_MAX_AGE
             LOG.info(
                 f"[{markup_color}]Token expires: {expiration_time} (in {readable_timedelta(lifetime)})[/{markup_color}]"
             )
-        else:
-            LOG.info(f"[{markup_color}]Token expired: {expiration_time}[/{markup_color}]")
-
-        if consignee:
```
MatthiasZepper, line 322: I am not happy that you dropped the whole consignee functionality. Even though I certainly have not the last say on this, I think it was very welcome from a UX perspective. 

Also mind that [Anands PR](https://github.com/ScilifelabDataCentre/dds_cli/pull/244) relies on this funcitonality.

dds_cli/user.py
```diff
@@ -312,40 +309,30 @@ def token_report(self, token_metadata=None):
...
+        (based on the token file) for the token's age and expiration time in UTC."""
+
+        expiration_time = get_token_expiration_time(token=token)
+
+        # Try to use the Expiration Time (exp) in the token jose header,
+        # otherwise fall back to calculation based on the token file
+        if expiration_time:
+            utc_tz = pytz.timezone("UTC")
+            expiration_time = datetime.datetime.fromtimestamp(expiration_time, tz=utc_tz)
+            age = dds_cli.TOKEN_MAX_AGE - (expiration_time - datetime.datetime.utcnow())
```
MatthiasZepper, line 332: If you do not want the time when the token was issued to the token itself for length constraints, then why don't you use the `modification time` here, too? 

I think this is a better approximation than a configured the `dds_cli.TOKEN_MAX_AGE` configured value, which I would prefer to only have as a fallback if possible (but I can also be convinced otherwise - more of a gut feeling)

dds_cli/user.py
```diff
@@ -312,40 +309,30 @@ def token_report(self, token_metadata=None):
...
         else:
-            expiration_time = issued_at + dds_cli.TOKEN_MAX_AGE
-
-        # lifetime is returned as a negative value / countdown
-        lifetime = datetime.datetime.utcnow().replace(tzinfo=utc_tz) - expiration_time
+            modification_time = datetime.datetime.fromtimestamp(os.path.getmtime(self.token_file))
```
MatthiasZepper, line 334: Mind that in the current version I correct for the local timezone, since `os.path.getmtime` returns the time in the local system time. So you need to have a time zone correction here!

I have tried it out and without using `tzlocal.get_localzone()` there was an offset of 1h due to Stockholm time. Should we ever have users located in Japan or the US, that will of course be more significant.

dds_cli/user.py
```diff
@@ -312,40 +309,30 @@ def token_report(self, token_metadata=None):
         LOG.info(f"[{markup_color}]{sign}  {message} {sign} [/{markup_color}]")
         LOG.info(f"[{markup_color}]Token age: {readable_timedelta(age)}[/{markup_color}]")
 
-        if lifetime.total_seconds() < 0:
+        if age > dds_cli.TOKEN_MAX_AGE:
```
MatthiasZepper, line 312: I would prefer to go more lightly on relying on `dds_cli.TOKEN_MAX_AGE`, such that the lifetime is calculated once in `__token_dates()` and than only used throughout the conditions.

MatthiasZepper (changes requested):
Most comments are mostly gut feeling and soft preferences for a different approach.

However, I think two things need to be addressed still:

- Timezone conversion for `os.path.getmtime()`:
https://github.com/ScilifelabDataCentre/dds_cli/blob/9180ea184284efcab48942b210f70873a3ddcddd/dds_cli/user.py#L334-L336

- What to do about the issued to username/consignee, also with regard to [Anands open PR that relies on this funcitonality](https://github.com/ScilifelabDataCentre/dds_cli/pull/244)? 
https://github.com/ScilifelabDataCentre/dds_cli/blob/1dee5d61104a9f3fe6c135a8fed1cd38d9442069/dds_cli/user.py#L322-L324

worukan:
> * What to do about the issued to username/consignee, also with regard to [Anands open PR that relies on this funcitonality](https://github.com/ScilifelabDataCentre/dds_cli/pull/244)?
>   https://github.com/ScilifelabDataCentre/dds_cli/blob/1dee5d61104a9f3fe6c135a8fed1cd38d9442069/dds_cli/user.py#L322-L324

As I wrote in Anand's PR:
This can be handled by showing the username that the user has entered to login. There is no need to add this to the plaintext part of the token and extract it from there. So if cli can get the final token to use the endpoints, the entered username can be displayed as logged in user.

dds_cli/user.py
```diff
@@ -258,48 +252,51 @@ def check_token_file_permissions(self):
...
+
         Returns True if the token has expired, False otherwise.
         """
-        token_dates = self.__token_dates(token_metadata=token_metadata)
+        age, expiration_time = self.__token_dates(token=token)
 
-        LOG.debug(f"Token file age: {readable_timedelta(token_dates['age'])}")
-        if token_dates["lft"].total_seconds() >= 0:
+        LOG.debug(f"Token file age: {readable_timedelta(age)}")
+        if age > dds_cli.TOKEN_MAX_AGE:
```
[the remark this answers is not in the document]
worukan: It comes from there.

dds_cli/user.py
```diff
@@ -258,48 +252,51 @@ def check_token_file_permissions(self):
...
             LOG.debug("Token has expired. Now deleting it and fetching new token.")
             self.delete_token()
             return True
-        elif dds_cli.TOKEN_WARNING_AGE * dds_cli.TOKEN_MAX_AGE <= token_dates["age"]:
+        elif age > dds_cli.TOKEN_WARNING_AGE:
+            lifetime = age - dds_cli.TOKEN_MAX_AGE
```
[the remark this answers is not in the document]
worukan: It is a simple calculation and not widely used. The values all depend on each other, so I don't see a reason to complicate it.

dds_cli/user.py
```diff
@@ -258,48 +252,51 @@ def check_token_file_permissions(self):
...
-        lifetime = token_dates["lft"]
-        expiration_time = token_dates["exp"]
+        age, expiration_time = self.__token_dates(token=token)
         # display expiration time in local time
         expiration_time = expiration_time.astimezone(tz=tzlocal.get_localzone()).strftime(
             "on %d %B %Y at %H:%Mh"
         )
 
-        if lifetime.total_seconds() >= 0:
+        if age > dds_cli.TOKEN_MAX_AGE:
```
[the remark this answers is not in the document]
worukan: age and expiration time depend on each other, so it is taken into account.

dds_cli/user.py
worukan: Already replied to this

dds_cli/user.py
worukan: modification time depends on token file, where you comment is a token header. They are unrelated. If exp in token header, we use that. Issue time is not a required claim and we don't need it at least for now. Users need to deal with the expiration time, because after that the token will be expired. Knowing the issue time of a token probably does not help a user.

dds_cli/user.py
worukan: If the common goal is to get rid of dds_cli.TOKEN_MAX_AGE, then let me know. I was about to delete it anyway. @inaod568 @alneberg

dds_cli/user.py
MatthiasZepper: I agree. I kept the `TOKEN_MAX_AGE` and `TOKEN_WARNING_AGE` because those were initially introduced by Johannes. Shouldn't we then not replace both with a `REMAINING_LIFETIME_WARNING` timedelta? Such that we don't bother at all how old the token is, but just warn once there are only X hours of lifetime left until it expires?

worukan: I think a notification that warns that the token will soon expire is good enough. For example, this is how the GitHub token works. I generate a token and forget about it until GitHub tells me that there are a few days left until it expires, so I go and regenerate it. We don't need to bother with age and etc. The expiration time gives us the deadline.

MatthiasZepper: Yes, sounds good. And the warning limit for soon could be a locally configured value like `REMAINING_LIFETIME_WARNING_PERIOD` or such.

dds_cli/user.py
alneberg: Yes,  I think we can get rid of that one if we can read the expiration time directly from the token. I believe Matthias kept it as a backup in case the age wouldn't be readable for some reason.

dds_cli/user.py
alneberg: I agree, warning when x hours remain of the token is good.

dds_cli/user.py
MatthiasZepper: Yes, indeed that was the initial, nowadays obsolete consideration. When I first wrote that code, I still hat the issue with the _PyJWT Invalid Payload Padding_ exception being raised, so I assumed that reading the header might occasionally fail.

inaod568 (approved):
👍🏻

MatthiasZepper (approved):
Very nice! 

The "_less than a minute_" is nifty - that would even work for a token age. I think, I pondered about 20min on a suitable on fits all solution before introducing the "_recent_" "_in a short while_" if/else statement instead.

PR #261, opened 2022-02-25
added unit and changed url

inaod568:
Before submitting a pr:
- [x] Tests passing
- [x] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/account_manager.py
+++ b/dds_cli/account_manager.py
@@ -48,7 +48,7 @@ class AccountManager(dds_cli.base.DDSBaseClass):
         if self.method not in ["add", "delete", "revoke"]:
             raise dds_cli.exceptions.AuthenticationError(f"Unauthorized method: '{self.method}'")
 
-    def add_user(self, email, role, project, unit, no_mail=False):
+    def add_user(self, email, role, project, unit=None, no_mail=False):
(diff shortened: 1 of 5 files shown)


dds_cli/account_manager.py
```diff
@@ -48,10 +48,10 @@ def __init__(
         if self.method not in ["add", "delete", "revoke"]:
             raise dds_cli.exceptions.AuthenticationError(f"Unauthorized method: '{self.method}'")
 
-    def add_user(self, email, role, project, no_mail=False):
+    def add_user(self, email, role, project, unit, no_mail=False):
```
MatthiasZepper, line 51: I think you should specify a default here, possibly an empty string `unit=""`. The reason is, that also the CLI function `main.py:grant_project_access` also calls the `add_user` method: 

https://github.com/ScilifelabDataCentre/dds_cli/blob/b38634da3f865ccdb8522ad7e6b2e1c9cb3ca07a/dds_cli/__main__.py#L949

MatthiasZepper (approved):
One fix is still required, though.

dds_cli/account_manager.py
inaod568: Fixed!

aanil (approved):
👍

PR #294, opened 2022-03-01
set the upload chunk size to larger

inaod568:
This is a part of the `Qutoa Exceeded` issue. Turns out it's not caused by this but may be a good change anyway. Not a priority atm though.

- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/data_putter.py
+++ b/dds_cli/data_putter.py
@@ -12,6 +12,7 @@ import pathlib
 
 # Installed
 import boto3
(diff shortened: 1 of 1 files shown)


inaod568:
Need to benchmark at some point probably, but this does not help with the original issue: QuotaExceeded. The error only depends on the used amount of data, not the number of requests etc.

PR #329, opened 2022-03-02
Token path argument

alneberg:
A new argument `--token-path` is added and used as such:

```
dds --token-path my_custom_token auth login
dds --token-path my_custom_token ls
```

I also moved the username printing for the logo inside the `dds_main`. As a consequence, it will now not be printed if you run `dds --help` or `dds --version`, but otherwise it should be printed. 

Also, if the username is not found, for example if no token is found, it's not printed at all. This can of course be adjusted however we want to.

Feel free to squash merge this as the commits became a bit messy.

- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG


dds_cli/user.py
```diff
@@ -206,8 +215,11 @@ def get_user_name_if_logged_in():
 class TokenFile:
     """A class to manage the saved token."""
 
-    def __init__(self):
-        self.token_file = dds_cli.TOKEN_FILE
+    def __init__(self, token_path=None):
+        if token_path is None:
```
inaod568, line 219: (Minor comment: You don't need the `is None` here)

inaod568 (approved):
Nice! Looks good and works!

aanil (approved):
👍

PR #364, opened 2022-03-08
Escape values that are printed using rich

talavis:
Escape values that are printed using rich, e.g. filenames containing `[]`.

This fixes all cases I could see during normal use. There are (most likely) places that I have missed for e.g. json or verbose output as well as error messages that I didn't trigger.

- [X] Tests passing
- [X] Black formatting
- [X] Rebase/merge the `dev` branch
- [ ] Note in the CHANGELOG


dds_cli/account_manager.py
```diff
@@ -194,6 +194,9 @@ def get_user_info(self):
 
             # Get response
             response_json = response.json()
+            for field in response_json["info"]:
+                if isinstance(response_json["info"][field], str):
+                    response_json["info"][field] = rich.markup.escape(response_json["info"][field])
```
inaod568, line 199: There's a risk that this will produce a crash. If the `"info"` doesn't exist in the response_json. 

I'd move it out of the try except and add 
```
if info:
    for field in info:
        if instance(response_json["info"][field], str): 
            response_json["info"][field] = rich.markup.escape(response_json["info"][field])
```
under 
```
info = response_json.get("info")
```
on row 216

inaod568:
```suggestion
```

inaod568 (changes requested):

dds_cli/account_manager.py
talavis: I can't move it if we want correct values in `LOG.debug`.

inaod568 (approved):
Works well! 👏🏻

PR #377, opened 2022-03-10
utils.calculate_magnitude can now handle vastly different project sizes

MatthiasZepper:
- [x] Tests passing
- [x] Black formatting
- [x] Note in the CHANGELOG

Handle the edge case that a user has two projects with a extremely different size available to them in the utils.calculate_magnitude function

- "auto": Use a consistent unit for all projects unless the variation is too big, in SI units (base 10). 
- "auto-iec": Use consistent unit for all projects unless the variation is too big, in IEC units (base 2). 
- "const": Use consistent unit for all projects, in SI units (base 10). 
- "const-iec": Use consistent unit for all projects, in IEC units (base 2). 
- "off": Display the project sizes in bytes, no further unit simplification."

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -1307,14 +1307,13 @@ def get_data(
 # Options
 @project_option(required=True)
 @folder_option(help_message="List contents in this project folder.")
-@unit_prefix_option
 # Flags
 @json_flag(help_message="Output in JSON format.")
 @size_flag(help_message="Show size of project contents.")
 @tree_flag(help_message="Display the entire project(s) directory tree.")
 @users_flag(help_message="Display users associated with a project(Requires a project id).")
 @click.pass_context
-def list_data(ctx, project, folder, unitprefix, json, size, tree, users):
+def list_data(ctx, project, folder, json, size, tree, users):
     """List project contents.
 
     Same as dds ls [PROJECT ID].
@@ -1323,7 +1322,6 @@ def list_data(ctx, project, folder, unitprefix, json, size, tree, users):
         list_projects_and_contents,
         project=project,
         folder=folder,
-        unitprefix=unitprefix,
         size=size,
         tree=tree,
         users=users,
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -46,6 +46,7 @@ from dds_cli.options import (
     source_option,
     source_path_file_option,
     token_path_option,
+    unit_prefix_option,
     break_on_fail_flag,
     json_flag,
     nomail_flag,
@@ -160,6 +161,7 @@ def dds_main(click_ctx, verbose, log_file, no_prompt, token_path):
 # Options
 @project_option(required=False)
 @sort_projects_option()
+@unit_prefix_option()
 @folder_option(help_message="List contents of this project folder.")
 # Flags
 @json_flag(help_message="Output in JSON format.")
@@ -170,7 +172,7 @@ def dds_main(click_ctx, verbose, log_file, no_prompt, token_path):
 @click.option("--projects", "-lp", is_flag=True, help="List all project connected to your account.")
 @click.pass_obj
 def list_projects_and_contents(
-    click_ctx, project, folder, sort, json, size, tree, usage, users, projects
+    click_ctx, project, folder, sort, unitprefix, json, size, tree, usage, users, projects
 ):
     """List the projects you have access to or the project contents.
 
@@ -189,7 +191,7 @@ def list_projects_and_contents(
                 json=json,
                 token_path=click_ctx.get("TOKEN_PATH"),
             ) as lister:
-                projects = lister.list_projects(sort_by=sort)
+                projects = lister.list_projects(sort_by=sort, unitprefix=unitprefix)
                 if json:
                     dds_cli.utils.console.print_json(data=projects)
                 else:
@@ -671,16 +673,17 @@ def project_group_command(_):
 @project_group_command.command(name="ls")
 # Options
 @sort_projects_option()
+@unit_prefix_option()
 # Flags
 @usage_flag(help_message="Show the usage for available projects, in GBHours and cost.")
 @json_flag(help_message="Output project list as json.")  # users, json, tree
 @click.pass_context
-def list_projects(ctx, json, sort, usage):
+def list_projects(ctx, json, sort, unitprefix, usage):
     """List all projects you have access to in the DDS.
 
     Calls the `dds ls` function.
     """
-    ctx.invoke(list_projects_and_contents, json=json, sort=sort, usage=usage)
+    ctx.invoke(list_projects_and_contents, json=json, sort=sort, unitprefix=unitprefix, usage=usage)
 
 
 # -- dds project create -- #
@@ -1304,13 +1307,14 @@ def get_data(
 # Options
 @project_option(required=True)
 @folder_option(help_message="List contents in this project folder.")
+@unit_prefix_option
 # Flags
 @json_flag(help_message="Output in JSON format.")
 @size_flag(help_message="Show size of project contents.")
 @tree_flag(help_message="Display the entire project(s) directory tree.")
 @users_flag(help_message="Display users associated with a project(Requires a project id).")
 @click.pass_context
-def list_data(ctx, project, folder, json, size, tree, users):
+def list_data(ctx, project, folder, unitprefix, json, size, tree, users):
     """List project contents.
 
     Same as dds ls [PROJECT ID].
@@ -1319,6 +1323,7 @@ def list_data(ctx, project, folder, json, size, tree, users):
         list_projects_and_contents,
         project=project,
         folder=folder,
+        unitprefix=unitprefix,
         size=size,
         tree=tree,
         users=users,
--- a/dds_cli/data_lister.py
+++ b/dds_cli/data_lister.py
@@ -506,8 +506,8 @@ class DataLister(base.DDSBaseClass):
                 "footer": dds_cli.utils.format_api_response(
                     total_size,
                     key="Size",
-                    iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
-                    skip=True if unitprefix == "off" else False,
+                    iec_standard=unitprefix in ["auto-iec", "const-iec"],
+                    skip=unitprefix == "off",
                 ),
                 "overflow": "ellipsis",
             }
@@ -522,8 +522,8 @@ class DataLister(base.DDSBaseClass):
                         "footer": dds_cli.utils.format_api_response(
                             usage_info["usage"],
                             key="Usage",
-                            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
-                            skip=True if unitprefix == "off" else False,
+                            iec_standard=unitprefix in ["auto-iec", "const-iec"],
+                            skip=unitprefix == "off",
                         ),
                         "overflow": "ellipsis",
                     },
@@ -533,8 +533,8 @@ class DataLister(base.DDSBaseClass):
                         "footer": dds_cli.utils.format_api_response(
                             usage_info["cost"],
                             key="Cost",
-                            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
-                            skip=True if unitprefix == "off" else False,
+                            iec_standard=unitprefix in ["auto-iec", "const-iec"],
+                            skip=unitprefix == "off",
                         ),
                         "overflow": "ellipsis",
                     },
@@ -584,8 +584,8 @@ class DataLister(base.DDSBaseClass):
         magnitudes = dds_cli.utils.calculate_magnitude(
             sorted_projects,
             column_formatting.keys(),
-            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
-            force=True if unitprefix in ["const", "const-iec"] else False,
+            iec_standard=unitprefix in ["auto-iec", "const-iec"],
+            force=unitprefix in ["const", "const-iec"],
         )
 
         # Add all column values for each row to table
--- a/dds_cli/data_lister.py
+++ b/dds_cli/data_lister.py
@@ -597,8 +597,8 @@ class DataLister(base.DDSBaseClass):
                             proj[i],
                             i,
                             magnitudes[i],
-                            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
-                            skip=True if unitprefix == "off" else False,
+                            iec_standard=unitprefix in ["auto-iec", "const-iec"],
+                            skip=unitprefix == "off",
                         )
                     )
                     for i in column_formatting
--- a/dds_cli/data_lister.py
+++ b/dds_cli/data_lister.py
@@ -74,7 +74,7 @@ class DataLister(base.DDSBaseClass):
 
     # Public methods ########################### Public methods #
 
-    def list_projects(self, sort_by="Updated"):
+    def list_projects(self, sort_by="Updated", unitprefix="auto"):
         """Get a list of project(s) the user is involved in."""
         # Get projects from API
         try:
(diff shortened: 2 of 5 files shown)


dds_cli/utils.py
```diff
@@ -53,13 +53,19 @@ def calculate_magnitude(projects, keys, iec_standard=False):
             # exclude values smaller than base, such that empty projects don't interfer with
             # the calculation ensures that a minimum can be calculated if no val is larger than base
             minimum = (lambda x: min(x) if x else 1)([val for val in values if val >= base])
-            mag = 0
+            maximum = (lambda x: max(x) if x else 1)([val for val in values if val >= base])
+            magmin = magmax = 0
```
ewels, line 57: General point - I'm always scared of doing this because of Python's mutability, the same thing that @talavis was just talking about in Slack about function defaults.

It's fine here because we're setting to an int which is immutable, however if you did `x = y = {}` then `x` would change if you changed `y`, because dicts are mutable.

https://note.nkmk.me/en/python-multi-variables-values/

> Be careful when assigning mutable objects such as `list` or `dict` instead of immutable objects such as `int`, `float`, or `str`.

ewels (commented):
Haven't tested, but code looks good 👍🏻

dds_cli/utils.py
```diff
@@ -53,13 +53,19 @@ def calculate_magnitude(projects, keys, iec_standard=False):
             # exclude values smaller than base, such that empty projects don't interfer with
             # the calculation ensures that a minimum can be calculated if no val is larger than base
             minimum = (lambda x: min(x) if x else 1)([val for val in values if val >= base])
-            mag = 0
+            maximum = (lambda x: max(x) if x else 1)([val for val in values if val >= base])
```
talavis, line 56:
```suggestion
minimum, maximum = (lambda x: min(x), max(x) if x else (1,1))([val for val in values if val >= base])
```

dds_cli/utils.py
MatthiasZepper: Thanks for the explanation. I will change that. 

It is the first time that I wrote something like this. Given Pythons amazing brevity, I believed that would be the "Pythonic" way to assign one value to two variables.

I thought, I had seen such a notation somewhere in our code before, but I must have mixed it up with this: 

https://github.com/ScilifelabDataCentre/dds_web/blob/4549858baef0eac95b0d7d0d9a3803e074af4e22/dds_web/api/user.py#L287

dds_cli/utils.py
MatthiasZepper: Thank you very much for the suggestion. Indeed much cleaner! I was eager to prove my point that it is a quick fix in the code that I choose the code duplication approach over carefully thinking it through...

MatthiasZepper: I changed it accordingly. I couldn't directly commit your suggestion though, because suggestions on deleted lines are apparently not supported. 

But thanks a lot: Very elegant and much cleaner!

dds_cli/utils.py
ewels: Yes exactly, it is quite nice any Pythonic. Just given our recent conversation about mutability causing problems I thought it was a good point to bring up for future coding :)

dds_cli/__main__.py
```diff
@@ -1286,13 +1289,14 @@ def get_data(
 # Options
 @project_option(required=True)
 @folder_option(help_message="List contents in this project folder.")
+@unit_prefix_option
```
inaod568, line 1292: I don't think you need this here? 
The unit prefix option is only for the project listing right?

dds_cli/data_lister.py
```diff
@@ -501,7 +503,12 @@ def __format_project_columns(self, total_size=None, usage_info=None):
             column_formatting["Size"] = {
                 "justify": "right",
                 "style": default_format.get("style"),
-                "footer": dds_cli.utils.format_api_response(total_size, key="Size"),
+                "footer": dds_cli.utils.format_api_response(
+                    total_size,
+                    key="Size",
+                    iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
+                    skip=True if unitprefix == "off" else False,
```
inaod568, line 510:
```suggestion
                    skip = unitprefix == "off",
```

dds_cli/data_lister.py
```diff
@@ -501,7 +503,12 @@ def __format_project_columns(self, total_size=None, usage_info=None):
             column_formatting["Size"] = {
                 "justify": "right",
                 "style": default_format.get("style"),
-                "footer": dds_cli.utils.format_api_response(total_size, key="Size"),
+                "footer": dds_cli.utils.format_api_response(
+                    total_size,
+                    key="Size",
+                    iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
```
inaod568, line 509:
```suggestion
                    iec_standard=unitprefix in ["auto-iec", "const-iec"],
```

dds_cli/data_lister.py
```diff
@@ -513,25 +520,33 @@ def __format_project_columns(self, total_size=None, usage_info=None):
                         "justify": "right",
                         "style": default_format.get("style"),
                         "footer": dds_cli.utils.format_api_response(
-                            usage_info["usage"], key="Usage"
+                            usage_info["usage"],
+                            key="Usage",
+                            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
```
inaod568, line 525:
```suggestion
                            iec_standard=unitprefix in ["auto-iec", "const-iec"],
```

dds_cli/data_lister.py
```diff
@@ -513,25 +520,33 @@ def __format_project_columns(self, total_size=None, usage_info=None):
...
+                            skip=True if unitprefix == "off" else False,
```
inaod568, line 526:
```suggestion
                            skip=unitprefix == "off",
```

dds_cli/data_lister.py
```diff
@@ -513,25 +520,33 @@ def __format_project_columns(self, total_size=None, usage_info=None):
...
                     },
                     "Cost": {
                         "justify": "right",
                         "style": default_format.get("style"),
-                        "footer": dds_cli.utils.format_api_response(usage_info["cost"], key="Cost"),
+                        "footer": dds_cli.utils.format_api_response(
+                            usage_info["cost"],
+                            key="Cost",
+                            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
+                            skip=True if unitprefix == "off" else False,
```
inaod568, line 537:
```suggestion
                            iec_standard=unitprefix in ["auto-iec", "const-iec"],
                            skip=unitprefix == "off",
```

dds_cli/data_lister.py
```diff
@@ -559,13 +574,26 @@ def __print_project_table(self, sorted_projects, usage_info, total_size):
             )
 
         # calculate the magnitudes for keeping the unit prefix constant across all projects
-        magnitudes = dds_cli.utils.calculate_magnitude(sorted_projects, column_formatting.keys())
+        magnitudes = dds_cli.utils.calculate_magnitude(
+            sorted_projects,
+            column_formatting.keys(),
+            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
+            force=True if unitprefix in ["const", "const-iec"] else False,
```
inaod568, line 581:
```suggestion
            iec_standard=unitprefix in ["auto-iec", "const-iec"],
            force=unitprefix in ["const", "const-iec"],
```

dds_cli/data_lister.py
```diff
@@ -559,13 +574,26 @@ def __print_project_table(self, sorted_projects, usage_info, total_size):
...
             table.add_row(
                 *[
-                    escape(dds_cli.utils.format_api_response(proj[i], i, magnitudes[i]))
+                    escape(
+                        dds_cli.utils.format_api_response(
+                            proj[i],
+                            i,
+                            magnitudes[i],
+                            iec_standard=True if unitprefix in ["auto-iec", "const-iec"] else False,
+                            skip=True if unitprefix == "off" else False,
```
inaod568, line 594:
```suggestion
                            iec_standard=unitprefix in ["auto-iec", "const-iec"],
                            skip=unitprefix == "off",
```

dds_cli/options.py
```diff
@@ -187,6 +187,35 @@ def token_path_option(
...
 
+def unit_prefix_option(
+    long="--unit-prefixes",
+    short="-up",
+    name="unitprefix",
+    choices=["auto", "auto-iec", "const", "const-iec", "off"],
+    case_sensitive=False,
+    default="auto",
+    required=False,
+    help_message="Use unit prefixes in order to reduce the number of digits. For Bytes, the SI prefixes can be replaced by powers of 1024 (IEC prefixes).",
```
inaod568, line 198:
```suggestion
    help_message=("Unit formatting in the Size column. "auto": Use a consistent unit for all projects unless the variation is too big, in SI units (base 10). "auto-iec": Use consistent unit for all projects unless the variation is too big, in IEC units (base 2). "const": Use consistent unit for all projects, in SI units (base 10). "const-iec": Use consistent unit for all projects, in IEC units (base 2). "off": Display the project sizes in bytes, no further unit simplification.
```

MatthiasZepper: So are you fine with such a comprehensive help message? I was thinking those should be more brief?

inaod568: Yeah. Tried now and we should keep it brief. Will change my suggestion.

inaod568: Done.

inaod568:
Screenshot 2022-03-15 at 12 59 45

* `someunit00001`: 
    * 3 files: 90 bytes, 263 bytes and 201 MB
* `someunit00002`:
    * 1 file: 90 bytes
* `someunit00003`: 
    * 1 file: 263 bytes
* `someunit00004`:
    * 1 file: 201 MB

1. We shouldn't have the decimals: `201.00` as the auto
2. We shouldn't display them as 0 if they actually contain something

So what we need to do here is fix so that the formats are correct, we don't need to fix consistent etc formatting at the moment. For now all we need is that the numbers are correct, and that we can display individual, exact formats.

dds_cli/options.py
MatthiasZepper: I am not strictly opposed to changing the text, but I find this suggestion unfortunately not helpful:

- It is not the unit formatting, but the use of unit metric prefixes that is configured.
- It is not exclusively the `Size` column that is affected, but `Size`, `Usage` and `Cost `.
- _See documentation for more information_ is relatively self-evident. I would presume that those help texts are in general read mostly by people who per se know how the system works, but the meaning of one particular config option has escaped them at the moment. 

What do you think of:

 _Configure the output's metric unit prefixes: Consistent scaling is enforced with const, auto optimizes for fewer digits. The -iec modifier allows to use binary instead of decadic prefixes for bytes_.
 
 ?

inaod568: Ok fair enough.
I don't think most people understand what "Configure the output's metric unit prefixes" means though, for example. So I think we need a more user friendly text.

inaod568:
> > ```
> > 1. We shouldn't have the decimals: `201.00` as the auto
> > ```
> 
> What should be the default then? Just 201?
> 
> (_The reason to print the trailing 0s here was to ensure consistent alignment of the decimal dot for all numbers in the column.)_

What I mean is the projects are said to be `201.00` are actually `201.42` for example. The `00` decimals are definitely correct and ok if the file is exactly the amount, but in this case they're not. 


> > ```
> > 2. We shouldn't display them as 0 if they actually contain something
> > ```
> 
> How should this be done? Both, 90 bytes and 263 bytes are smaller than the base of the decimal multiplicative prefixes (1000 respectively 1024 bytes). Thus, displaying them is only possible if scaling is suppressed by `off`. I am hesitant to add more trailing decimals than 2 will only cause confusion with bigger values?

This is how I think it should work: 
- Default should be that the exact values are shown, in SI
- Option to display the exact values, in IEC
- Option to set the values to a consistent SI format
- Option to set the values to a consistent IEC format
- Option to only display in bytes

This is what I meant last week though, with it not being a priority and that we shouldn't spend time on it if it's not very fast. Basically all we need atm is that values that are shown are correct (so basically the first point above).

ewels:
> The `00` decimals are definitely correct and ok if the file is exactly the amount, but in this case they're not.

I bet something is being treated as an `int` instead of a `float` somewhere. I get this behaviour a lot with MultiQC.

dds_cli/options.py
MatthiasZepper: Ok, how about: "_Adjust units to improve readability, either consistently with 'const' or adaptively with 'auto'. Prefixes are either decadic (Kilo-,Mega-,Giga-) by default or binary (Kibi-,Mebi-,Gibi-) when specifying '-iec'._"

alneberg: My try:
__Configure_ the output's unit prefixes: const results in identical unit prefixes between rows, auto optimizes for fewer digits. The -iec modifier allows to use binary instead of decadic prefixes for bytes._

Please correct me if this is in fact incorrect though!

inaod568: > My try: __Configure_ the output's unit prefixes: const results in identical unit prefixes between rows, auto optimizes for fewer digits. The -iec modifier allows to use binary instead of decadic prefixes for bytes._
> 
> Please correct me if this is in fact incorrect though!

Yes, I like this one. Lets use this 👍🏻  @MatthiasZepper I'm assuming the thumbs up from your side is that it's correct so you can ignore my suggestion and just use when Johannes said here, ok?

MatthiasZepper: Yes, the thumbs up referred to the correctness. However, does this suggestion sufficiently address your concern that people do not know what "_Configure the output's metric unit prefixes"_ means? Except for the word "_metric_" it is pretty much similar or am I missing something. 

I tried to include examples (Kilo,Mega,Giga) to exemplify in case someone doesn't know what a unit prefix is and stepped back from detail description of the options to just call the formatting _adaptive_ and _consistent_. Does any of those sound like an additional improvement or should be exactly go with Johannes proposal?

inaod568: @MatthiasZepper No it doesn't. But I mean that this is clearer at least. Having those examples that you mention would make it better.

So in that case
```suggestion
    help_message="Configure the output's unit prefixes (Kilo, Mega, Giga): const results in identical unit prefixes between rows, auto optimizes for fewer digits. The -iec modifier allows to use binary instead of decadic prefixes for bytes.",
```
?

MatthiasZepper: I have slightly rephrased it, but in general it would work. The length of both proposals with 222 respectively 225 characters might be an issue, though...

inaod568:
Screenshot 2022-03-18 at 12 00 30

`someunit00001` is actually 545 bytes.

@MatthiasZepper I'm gonna mark this as a draft for now. I like the idea of this, but it needs some more changes and we have some other tasks that need attending to more urgently. I can have a look again later on.

MatthiasZepper:
>  `someunit00001` is actually 545 bytes.

I am not surprised. [As I said](https://github.com/ScilifelabDataCentre/dds_cli/pull/377#issuecomment-1067962807) any value smaller than base is currently treated as insignificant and therefore displayed as 0. I don't see a problem there as it can always be skipped and the raw byte values be displayed with `off`.  

Correctly handling this edge case would bloat the functions logic significantly and also require to change the fixed two digits after 0 to a larger number adaptively to maintain alignment of the numbers. Therefore, we might then see displays like in the edge cases that you are envisioning:
```
200,000,000,000.00000000
000,000,000,000.00000034
```

I really don't see how this aids the sole purpose of the function: To improve readability and brevity.

inaod568:
> > `someunit00001` is actually 545 bytes.
> 
> I am not surprised. [As I said](https://github.com/ScilifelabDataCentre/dds_cli/pull/377#issuecomment-1067962807) any value smaller than base is currently treated as insignificant and therefore displayed as 0. I don't see a problem there as formatting can always be skipped and the raw byte values be displayed with `off`.
> 
> Correctly handling this edge case would bloat the functions logic significantly and also require to change the fixed two digits after 0 to a larger number adaptively to maintain alignment of the numbers. Therefore, we might then see displays like:
> 
> ```
> 200,000,000,000.00000000
> 000,000,000,000.00000545
> ```
> 
> I really don't see how this aids the sole purpose of the function: To improve readability and brevity.

This is not what I have been saying. The default should be to show the exact numbers, so in this case it should say "545 bytes" for example, without the zeroes. 

I have now fixed (https://github.com/ScilifelabDataCentre/dds_cli/pull/410) the project listing. This is the way I want the default to be, and this is what needs to be fixed now. As I have repeatedly said, additional formatting/improvements/giving the choice of what prefixes to display as you have done here would be great to have, but it is not a priority and I don't have time to have this discussion anymore, so this PR will have to wait. 

I'm sorry if you do not agree with what I have done now, but as the project leader I decide on which code is approved and merged.

PR #410, opened 2022-03-20
Fix size usage cost 

inaod568:
Currently the project listing functionality does not display the project sizes, usage and costs like I want them to. 
The default should be to display them in the exact form, not a common unit prefix and not any automatic recognition of the numbers. 

In the future, after getting the DDS into production, we can improve on this and add additional options like @MatthiasZepper has done in https://github.com/ScilifelabDataCentre/dds_cli/pull/377 for example. This however should not change the default that I explained above. 

This PR also fixes the bug where the size is not shown for the Researchusers. 

- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/utils.py
+++ b/dds_cli/utils.py
@@ -217,7 +217,7 @@ def format_api_response(response, key: str):
             if key == "Usage":
                 formatted_response += "H"
         elif key == "Cost":
-            formatted_response = "{:.1f}".format(response) if response >= 1.00 else str(0)
+            formatted_response = f"{response:.1f}" if response >= 1.00 else str(0) + " kr"
 
     return str(formatted_response)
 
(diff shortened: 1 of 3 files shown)


dds_cli/utils.py
```diff
@@ -12,13 +12,74 @@
...
 console = rich.console.Console()
 stderr_console = rich.console.Console(stderr=True)
 
+# Classes
+
+
+class HumanBytes:
+    """Format as human readable.
+
+    Partially copied from Stack Overflow: https://stackoverflow.com/a/63839503.
```
inaod568, line 29:
```suggestion
    Copied from Stack Overflow: https://stackoverflow.com/a/63839503.
```

PR #419, opened 2022-03-22
Release actions

MatthiasZepper:
Created two closely related Github Actions to automatically build the executable standalone versions of the CLI  (with help from @zishanmirza) and the documentations with Sphinx. The built files are attached to the Release.

The default `deploy-workflow.yml` is trigged by manual releases, whereas the `make-release.yml` creates the release automatically when a tag is detected that starts with a small v.


inaod568 (approved):
VSCode linting doesn't seem to like line 152 and below, because of "no value for parameter" but I'm ignoring that since it works! 
Gonna merge and try it out!

PR #422, opened 2022-03-23
Release actions (temp)

inaod568:
Same as the original by @MatthiasZepper.
This is temporary until there's a new PR.

The publishing to PyPi works, but then one of the actions fail because the "file already exists". So publishing the same is performed twice. 

Reverts ScilifelabDataCentre/dds_cli#421


inaod568:
@zishanmirza @MatthiasZepper 
Could you also add steps on how to do it (generate the executables with pyinstaller) manually here?

PR #463, opened 2022-04-20
Replace "requests.get" with the "request_get" function

zishanmirza:
This patch replaces "requests.get" with the "request_get" function in "utils.py".

Fixes DDS-1236.

- [X] Tests passing
- [X] Black formatting
- [X] Rebase/merge the `dev` branch
- [ - ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/base.py
+++ b/dds_cli/base.py
@@ -11,7 +11,6 @@ import pathlib
 
 # Installed
 import http
-import requests
 import simplejson
 
 # Own modules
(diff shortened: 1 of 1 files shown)


dds_cli/base.py
```diff
@@ -146,23 +145,12 @@ def __get_key(self, private: bool = False):
...
-            LOG.fatal(str(err))
-            raise SystemExit(
-                "Failed to get cloud information"
-                + (
-                    ": The database seems to be down."
-                    if isinstance(err, requests.exceptions.ConnectionError)
-                    else "."
-                )
-            ) from err
+        response = utils.request_get.get(
```
inaod568, line 148:
```suggestion
        response = utils.request_get(
```

inaod568 (changes requested):
Change in function name and also need to refactor the end of the function: https://github.com/ScilifelabDataCentre/dds_cli/pull/463/files#diff-51fea701513da58b2a0a4b204b195162b68d9f5f57e782277cc4365d4944a0f6R155-R175

```
if not response.ok:
```
etc is already checked by `request_get` and what's returned by the function is already in json.

dds_cli/base.py
```diff
@@ -146,23 +145,12 @@ def __get_key(self, private: bool = False):
...
+            DDSEndpoint.PROJ_PRIVATE if private else DDSEndpoint.PROJ_PUBLIC,
+            params={"project": self.project},
+            headers=self.token,
+            error_message="Failed to get cloud information: The database seems to be down.",
+        )
 
         if not response.ok:
```
valyo, line 155: this is handled inside the request_get and will fail here

valyo, line 155: this is handled inside the request_get and will fail here

valyo, line 155: the response check and getting the json are handled inside the request_get and will fail here anyways

dds_cli/base.py
```diff
@@ -146,37 +145,12 @@ def __get_key(self, private: bool = False):
...
-            raise SystemExit from err
+        project_public = utils.request_get(
+            DDSEndpoint.PROJ_PRIVATE if private else DDSEndpoint.PROJ_PUBLIC,
+            params={"project": self.project},
+            headers=self.token,
+            error_message="Failed to get cloud information: The database seems to be down.",
+        )
 
         if key_type not in project_public:
             utils.console.print(
```
inaod568, line 156: I just realised that the string here is missing an `f` before `"` so `{key_type}` will always be printed out in this case and not the actual value. Can you change that too please while you're at it?

inaod568 (changes requested):

dds_cli/base.py
```diff
@@ -146,37 +145,12 @@ def __get_key(self, private: bool = False):
...
-        try:
-            project_public = response.json()
-        except simplejson.JSONDecodeError as err:
-            LOG.fatal(str(err))
-            raise SystemExit from err
+        project_public = utils.request_get(
+            DDSEndpoint.PROJ_PRIVATE if private else DDSEndpoint.PROJ_PUBLIC,
+            params={"project": self.project},
+            headers=self.token,
+            error_message="Failed to get cloud information: The database seems to be down.",
```
inaod568, line 152: ": The database seems to be down." is defined in the `request_get` function. Also, this old message is a bit confusing and a bit misrepresenting, so change that row to this: 
```suggestion
            error_message="Failed to project key",
```

inaod568 (changes requested):

dds_cli/base.py
```diff
@@ -146,41 +145,16 @@ def __get_key(self, private: bool = False):
...
-        try:
-            project_public = response.json()
-        except simplejson.JSONDecodeError as err:
-            LOG.fatal(str(err))
-            raise SystemExit from err
+        project_public = utils.request_get(
+            DDSEndpoint.PROJ_PRIVATE if private else DDSEndpoint.PROJ_PUBLIC,
+            params={"project": self.project},
+            headers=self.token,
+            error_message="Failed to get project key.",
```
inaod568, line 152:
```suggestion
            error_message="Failed to get project key",
```

dds_cli/base.py
inaod568: @zishanmirza

PR #464, opened 2022-04-20
Refactor requests in account_manager.py

valyo:
- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/utils.py
+++ b/dds_cli/utils.py
@@ -147,7 +147,7 @@ def request_get(
     headers,
     params=None,
     json=None,
-    error_message=None,
+    error_message="API Request failed.",
     timeout=DDSEndpoint.TIMEOUT,
 ):
     """Perform get request."""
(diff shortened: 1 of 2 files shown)


dds_cli/utils.py
```diff
@@ -147,7 +147,7 @@ def request_get(
     headers,
     params=None,
     json=None,
-    error_message="API Request failed.",
+    error_message=None,
```
inaod568, line 150: We need to have some kind of default or require the user to input it (remove `=None`), otherwise there's a risk it will display `None: API Request failed` for example.

dds_cli/utils.py
```diff
@@ -167,7 +167,7 @@ def request_get(
                 + (
                     ": The database seems to be down."
                     if isinstance(err, requests.exceptions.ConnectionError)
-                    else "."
+                    else ": API Request failed."
```
inaod568, line 170:
```suggestion
                    else ": Unknown error."
```

inaod568 (changes requested):
Looks good! Just a couple of small changes needed.

dds_cli/utils.py
valyo: Yes indeed, I wanted to have it that way but I managed to fool myself that it should be None

inaod568 (approved):
Looks good!

PR #467, opened 2022-04-21
Refactor requests in data lister 

valyo:
- [x] Tests passing
- [x] Black formatting
- [ - ] Note in the CHANGELOG


inaod568 (changes requested):
I believe the tests fail because now we always have both `params` and `json` in the request and we're testing that it's only using `json`, so if you add that the request should call `params=None` the tests should work again

inaod568 (approved):

PR #478, opened 2022-05-02
Replace "requests.put/post" in "data_putter.py" with the "perform_request" function from "utils.py"

valyo:
- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/data_putter.py
+++ b/dds_cli/data_putter.py
@@ -13,7 +13,6 @@ import pathlib
 # Installed
 import boto3
 import botocore
-import requests
 from rich.markup import escape
 from rich.progress import Progress, SpinnerColumn, BarColumn
 import simplejson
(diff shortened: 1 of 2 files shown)


dds_cli/data_putter.py
```diff
@@ -399,28 +398,15 @@ def add_file_db(self, file):
...
-
-        return added_to_db, error
+        request_method = "put" if fileinfo["overwrite"] else "post"
+        response_json = dds_cli.utils.perform_request(
+            DDSEndpoint.FILE_NEW,
+            method=request_method,
+            params=params,
+            json=file_info,
+            headers=self.token,
+            error_message="Failed to add file '{file}' to database",
```
inaod568, line 408:
```suggestion
            error_message=f"Failed to add file '{file}' to database",
```

dds_cli/data_putter.py
```diff
@@ -399,28 +398,15 @@ def add_file_db(self, file):
...
+        )
+
+        added_to_db, message = (True, response_json)
+        return added_to_db, message
```
inaod568, line 412: In order to make sure that the upload of the rest of the files continue even if one fails we need to add a try except block here, so something like this I believe. 
```suggestion
        try:
            response_json = dds_cli.utils.perform_request(
                DDSEndpoint.FILE_NEW,
                method=request_method,
                params=params,
                json=file_info,
                headers=self.token,
                error_message="Failed to add file '{file}' to database",
            )
            added_to_db, message = (True, response_json)
        except (dds_cli.exceptions.ApiRequestError, dds_cli.exceptions.ApiResponseError, dds_cli.exceptions.DDSCLIException) as err:
            message = str(err)
            LOG.warning(message)

        return added_to_db, message
```

inaod568 (changes requested):

dds_cli/data_putter.py
valyo: I pushed the change but not sure how to simulate a file upload failure for testing right now. I continue with other tasks not to lose too much time with this testing

inaod568: Sure, that's fine, I can test and make changes in this if I see any that are needed.

inaod568 (approved):
This seems to work exactly like intended so 👍🏻

PR #484, opened 2022-05-04
fixes "IndexError: list index out of range" when listing files within project

valyo:
Fixes DDS-1262

- [ ] Tests passing
- [ ] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/data_lister.py
+++ b/dds_cli/data_lister.py
@@ -204,7 +204,7 @@ class DataLister(base.DDSBaseClass):
                 tabs_bf_format = th.TextHandler.format_tabs(
                     string_len=len(size), max_string_len=max_size, tab_len=2
                 )
-                line += f"{tabs_bf_format}{size.split()[1]}"
+                line += f"{tabs_bf_format}{size.split()[0]}"
             tree.add(line)
 
         # Print output to stdout


inaod568 (changes requested):
Tried it and it returns this instead: 
```
root@9db69c86ecb6:/code# dds ls -p someunit00001 --size
     ︵ 
 ︵ (  )   ︵ 
(  ) ) (  (  )   SciLifeLab Data Delivery System 
 ︶  (  ) ) (    http://dds_backend:5000/ 
      ︶ (  )    Version 1.0.3 
          ︶
Current user: unituser_1
INFO     Listing files for project 'someunit00001'                                                              
                                                                                                                
 Files / directories in project: someunit00001                                                                  
 └── tests/   49  49    
```
(the numbers next to each other)
Looked into it a bit more and it turns out this is actually an issue with that the API doesn't seem to be returning floats but strings, which means that the formatting isn't happening in the correct way in `def format_api_response`. To solve this bug, we should instead make sure that what's passed in to this function is in fact a number. For example, it can be solved in the `dds_web/api/files.py` module, on line 285 (`info.update({"size": x[1]})`) and line 296 (`info.update({"size": folder_size})`). 
If we change this to `info.update({"size": float(x[1])})` and `info.update({"size": float(folder_size)})` then it works again, without any changes to the CLI. 
But I'm not sure if there is a better way to fix it - have a look if we're maybe intentionally or unintentionally formatting the numbers as strings.

valyo:
Interesting, I will check how it looks with a number returned by the API and try to get a feeling if it is intentional or not

valyo:
It turns out the API is returning the files sizes as int and the folders sizes as decimal, and the function `format_api_response` does not handle the decimal properly. That's why this bug affects only the folders sizes listing. 
Changing the API to return float seems to solve the issue and I couldn't see any particular reason for the current return types.
Another thing is that `th.TextHandler.format_tabs` doesn't seem to handle the spacing properly:
```
 docs                                                                                                                                                                                  
 ├── Makefile            752.0B                                                                                                                                                        
 ├── _static/            1.9 MB                                                                                                                                                        
 ├── auth.rst            97.0 B                                                                                                                                                        
 ├── conf.py             2.2 KB                                                                                                                                                        
 ├── data.rst            97.0 B                                                                                                                                                        
 ├── index.rst           5.8 KB                                                                                                                                                        
 ├── installation.rst    5.5 KB                                                                                                                                                        
 ├── ls.rst              101.0B                                                                                                                                                        
 ├── main.rst            84.0 B                                                                                                                                                        
 ├── make.bat            795.0B                                                                                                                                                        
 ├── project.rst         106.0B                                                                                                                                                        
 ├── unit.rst            97.0 B                                                                                                                                                        
 └── user.rst            97.0 B 
```  

This is probably not a very urgent issue and we can either create a separate task, or maybe fix it directly if it is a quick fix.

inaod568:
> It turns out the API is returning the files sizes as int and the folders sizes as decimal, and the function `format_api_response` does not handle the decimal properly. That's why this bug affects only the folders sizes listing. Changing the API to return float seems to solve the issue and I couldn't see any particular reason for the current return types. Another thing is that `th.TextHandler.format_tabs` doesn't seem to handle the spacing properly:
> 
> ```
>  docs                                                                                                                                                                                  
>  ├── Makefile            752.0B                                                                                                                                                        
>  ├── _static/            1.9 MB                                                                                                                                                        
>  ├── auth.rst            97.0 B                                                                                                                                                        
>  ├── conf.py             2.2 KB                                                                                                                                                        
>  ├── data.rst            97.0 B                                                                                                                                                        
>  ├── index.rst           5.8 KB                                                                                                                                                        
>  ├── installation.rst    5.5 KB                                                                                                                                                        
>  ├── ls.rst              101.0B                                                                                                                                                        
>  ├── main.rst            84.0 B                                                                                                                                                        
>  ├── make.bat            795.0B                                                                                                                                                        
>  ├── project.rst         106.0B                                                                                                                                                        
>  ├── unit.rst            97.0 B                                                                                                                                                        
>  └── user.rst            97.0 B 
> ```
> 
> This is probably not a very urgent issue and we can either create a separate task, or maybe fix it directly if it is a quick fix.

Ah ok! Thanks for this explanation! 

Regarding the spacing - partly because the lines are formatted so that they're aligned to the right. I agree we may want to look at it at some point though, but yeah as you said not urgent and we can do this later. Closing this PR in favor of https://github.com/ScilifelabDataCentre/dds_web/pull/1162

PR #485, opened 2022-05-06
Project Creation Error

zishanmirza:
This patch updates the error messages for project creation.

Fixes DDS-1271.

- [X] Tests passing
- [x] Black formatting
- [ ] Note in the CHANGELOG

### DIFF
--- a/dds_cli/utils.py
+++ b/dds_cli/utils.py
@@ -12,7 +12,7 @@ from jwcrypto.jws import InvalidJWSObject
 from jwcrypto import jwt
 import http
 from rich.table import Table
-from typing import List, Union
+from typing import Dict, List, Union
 
 import dds_cli.exceptions
 from dds_cli import DDSEndpoint
@@ -190,6 +190,19 @@ def perform_request(
     # Check if response is ok.
     if not response.ok:
         message = error_message
+
+        if response.status_code == http.HTTPStatus.BAD_REQUEST:
+            if DDSEndpoint.CREATE_PROJ in endpoint:
+                message += f": {__project_creation_error(response_json)}"
+
+            raise dds_cli.exceptions.DDSCLIException(message=message)
+
+        if response.status_code == http.HTTPStatus.FORBIDDEN:
+            if DDSEndpoint.CREATE_PROJ in endpoint:
+                message += f": {response_json.get('message')}"
+
+            raise dds_cli.exceptions.DDSCLIException(message=message)
+
         if response.status_code == http.HTTPStatus.INTERNAL_SERVER_ERROR:
             raise dds_cli.exceptions.ApiResponseError(message=f"{message}: {response.reason}")
 
@@ -330,3 +343,22 @@ def delete_folder(folder):
         else:
             file_or_folder.unlink()
     folder.rmdir()
+
+
+def __project_creation_error(response_json: Dict) -> str:
+    message, title, description, pi, email = (
+        response_json.get("message"),
+        response_json.get("title"),
+        response_json.get("description"),
+        response_json.get("pi"),
+        response_json.get("email"),
+    )
+
+    messages: List = [message, title, description, pi, email]
+
+    error = next(message for message in messages if message)
+
+    if isinstance(error, List):
+        return error[0]
+
+    return error
--- a/tests/requirements-test.txt
+++ b/tests/requirements-test.txt
@@ -1,2 +1,4 @@
-pytest
+requests-mock==1.9.3
+pytest==7.1.2
+pytest-cov==3.0.0
 pyfakefs==4.5.5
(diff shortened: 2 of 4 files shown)


dds_cli/utils.py
```diff
@@ -330,3 +338,25 @@ def delete_folder(folder):
...
+
+    messages: List = [message, title, description, pi, email]
+
+    error = next(message for message in messages if message)
+
+    if isinstance(error, List):
+        return error[0]
+
+    if "Insufficient credentials" in error:
+        return "You do not have the required permissions to create a project."
```
inaod568, line 360: The code never reaches this part 
```
Current user: researchuser_1
ERROR    ⚠️ Failed to create project: Insufficient credentials ⚠️        
```
The "insufficient credentials" part is printed out from line 204.

I feel like this could be a separate issue though. I think you have solved the other error printouts in a good way so if you want to remove this part we can make a card to go through "Insufficient Credentials" errors in general within the entire CLI at a later time.

inaod568 (changes requested):

dds_cli/utils.py
```diff
@@ -190,6 +190,19 @@ def perform_request(
...
+            if DDSEndpoint.CREATE_PROJ in endpoint:
+                message += f": {__project_creation_error(response_json)}"
+
+            raise dds_cli.exceptions.DDSCLIException(message=message)
+
+        if response.status_code == http.HTTPStatus.FORBIDDEN:
+            if DDSEndpoint.USER_ADD in endpoint:
+                message += ": You do not have the required permissions to create a project."
+
+            raise dds_cli.exceptions.DDSCLIException(message=message)
```
inaod568, line 204: Do you mean `DDSEndpoint.CREATE_PROJ`? The `DDSEndpoint.USER_ADD` endpoint is not called from the CLI when we try to add users while creating a project. The project-creation endpoint handles that by calling the user-add endpoint but it's still the create project endpoint that's called here.

inaod568 (changes requested):

dds_cli/utils.py
```diff
@@ -190,6 +190,19 @@ def perform_request(
...
+            if DDSEndpoint.CREATE_PROJ in endpoint:
+                message += f": {__project_creation_error(response_json)}"
+
+            raise dds_cli.exceptions.DDSCLIException(message=message)
+
+        if response.status_code == http.HTTPStatus.FORBIDDEN:
+            if DDSEndpoint.CREATE_PROJ in endpoint:
+                message += ": You do not have the required permissions to create a project."
+
+            raise dds_cli.exceptions.DDSCLIException(message=message)
```
inaod568, line 204: I have created this now, will continue filling it in: https://github.com/ScilifelabDataCentre/dds_web/pull/1164 
This documents which status codes are returned by the API endpoints and when. 

Just adding this here because this also means that this if statement means that there will be an incorrect message when there are not enough unit admins. 

I think we should remove this check for now and add a better handling of FORBIDDEN at a later time.

inaod568 (changes requested):

valyo:
> @valyo Can you paste the lines 365-366, I don't see the checke? I think you're referring to the same thing as in my comment: [#485 (comment)](https://github.com/ScilifelabDataCentre/dds_cli/pull/485#discussion_r867985103)
> 
here is the function where those lines are:
 ```
def __project_creation_error(response_json: Dict) -> str:
    message, title, description, pi, email = (
        response_json.get("message"),
        response_json.get("title"),
        response_json.get("description"),
        response_json.get("pi"),
        response_json.get("email"),
    )

    messages: List = [message, title, description, pi, email]

    error = next(message for message in messages if message)

    if isinstance(error, List):
        return error[0]

    if "Insufficient credentials" in error:
        return "You do not have the required permissions to create a project."

    return error

```

**Comment by inaod568** [comment]
> > @valyo Can you paste the lines 365-366, I don't see the checke? I think you're referring to the same thing as in my comment: [#485 (comment)](https://github.com/ScilifelabDataCentre/dds_cli/pull/485#discussion_r867985103)
> 
> here is the function where those lines are:
> 
> ```
> def __project_creation_error(response_json: Dict) -> str:
>    message, title, description, pi, email = (
>        response_json.get("message"),
>        response_json.get("title"),
>        response_json.get("description"),
>        response_json.get("pi"),
>        response_json.get("email"),
>    )
> 
>    messages: List = [message, title, description, pi, email]
> 
>    error = next(message for message in messages if message)
> 
>    if isinstance(error, List):
>        return error[0]
> 
>    if "Insufficient credentials" in error:
>        return "You do not have the required permissions to create a project."
> 
>    return error
> ```

So you mean the 
```
   if "Insufficient credentials" in error:
       return "You do not have the required permissions to create a project."
```
part?
In that case yes, that part isn't reached. (Zishan has removed that part)

inaod568 (approved):
Seems to work as intended and I think I understand the tests, looks much better than it did before!

PR #540, opened 2022-09-30
Clarify the CLI installation instructions

valyo:

# Description
Improvement in the MacOS/Linux executable part. Added links to Apple documentation about allowing apps to run. I think an extensive how-to with screenshots is out of the scope of this documentation. 
Small changes in the Windows executable part too, but lack of Windows computer for testing makes it hard to achieve something better.

- [x] Fixes DDS-1311

## Type of change

- [x] Documentation

# Checklist:

## Checks

- [x] CodeQL passes
- [x] Tests
  - [x] The tests pass


valyo:
for Windows we could maybe add something like this https://www.mathworks.com/matlabcentral/answers/94933-how-do-i-edit-my-system-path-in-windows
since the explanation of how to use the history etc is not really good.

docs/installation.rst
```diff
@@ -65,19 +65,50 @@ PyPi
...
+   .. code-block:: bash
+
+      $ chmod +x dds-cli_ubuntu_x86_64   
+
+   or
+
+   .. code-block:: bash
+
+      $ chmod +x dds-cli_macos_x86_64   
+
```
inaod568, line 84: Looks good but I think we should probably clarify that _if_ the file is not executable and they don't have permission, they need to do this.

inaod568 (changes requested):

inaod568:
> for Windows we could maybe add something like this https://www.mathworks.com/matlabcentral/answers/94933-how-do-i-edit-my-system-path-in-windows since the explanation of how to use the history etc is not really good.

@valyo Sorry, not following. Why do they need to change the system path? Or was this an example of a step by step thing regarding how to get the path to the executable? Like you've added now? 
I think what you've added now is definitely good enough and the rest will be up to them / the units to help them with. If there are complaints / suggestions saying that we should add more info then we can discuss it again 👍🏻

docs/installation.rst
valyo: My impression is that the file is never executable, and also, if we write smth like _if the file is not executable_ then we need to add instructions about how to check that. Isn't it much simpler to just give them the _chmod_ command.

inaod568: Yeah true, just not sure we should encourage people to change permissions etc, but maybe I'm overly paranoid.

inaod568 (approved):
Looks good!

PR #541, opened 2022-10-04
Change message in exception when data already uploaded

inaod568:

# Description

1. An instance of `os.exit` changed to raise an exception. I believe this also fixes the issue with the `--source` option and the project remaining as `busy` @valyo: When no `source` or `source_path_file` was specified, a message was printed out and `os.exit` was used, which meant that the code never reached `except: self.cleanup_busy_status()` --> stuck in busy. Now when the code raises an exception instead, it does clean up the busy status.
2. Error message changed so that the user gets more information than "No data to upload" when the data has already been uploaded. 

- [x] Fixes DDS-1188

## Type of change

- [x] Documentation
- [x] Bug fix (non-breaking)

# Checklist:

## General

- [x] Code change

## Repository / Releases

- [x] Blocking PRs have been merged

## Checks

- [x] CodeQL passes
- [x] Tests
  - [x] The tests pass


valyo (approved):
Looks fine and works

PR #545, opened 2022-10-06
Raise exceptions instead of printing + os exit

inaod568:

# Description

These may also cause issues with the busy status as they are now, and we should anyway never be using `os._exit` instead of raising so we should change this anyway. 

## Type of change

- [x] Bug fix (non-breaking)

# Checklist:

## General

- [x] Code change
  - [x] Comments added, particularly in hard-to-understand areas

## Repository / Releases

## Checks

- [x] CodeQL passes
- [x] Tests
  - [x] The tests pass


valyo (approved):
I don't really know how to test it, but I don't see anything strange with the code changes

inaod568:
> I don't really know how to test it, but I don't see anything strange with the code changes

The way I tested it was by testing before and after without the if statements to see what happened with the messages and the busy status. You don't need to do this though, I've tried, so as long as you don't see anything strange 👍🏻

PR #563, opened 2022-10-28
Catch Timeout exception

inaod568:

# Description

When a requests takes too long, it times out, currently leading to the error message `ERROR Expecting value: line 1 column 1 (char 0)`. This PR catches the timeout exception and prints a more understandable message.

- [x] Fixes DDS-1402

## Type of change

- [x] Bug fix (non-breaking)

# Checklist:

## General

- [x] Code change

## Repository / Releases

## Checks

- [x] CodeQL passes
- [x] Tests
  - [x] The tests pass


valyo (approved):
Looks good

PR #564, opened 2022-10-28
Check that `dds data get` is used with either `--get-all` or `--source` and display appropriate message

valyo:

Currently if the user does not specify either of the options, it just returns a message saying “no data specified” which seems to confuse the users and result in support tickets.

Add a check for this and inform them in the CLI of the different options. 

- [x] Fixes [[link to issue / Jira issue ID]](https://scilifelab.atlassian.net/jira/software/projects/DDS/boards/13?selectedIssue=DDS-1414)

## Type of change

- [x] New feature (non-breaking)

# Checklist:

## General

- [x] Code change

## Checks

- [ ] CodeQL passes
- [x] Tests
  - [x] The tests pass

### DIFF
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -1607,7 +1607,7 @@ def get_data(
         sys.exit(1)
     elif not get_all and not (source or source_path_file):
         LOG.error(
-            "The command 'dds data get' reuires either the flag '--get-all' or some of the options '--source'/'--source-path-fail'."
+            "Specify either '--source'/'--source-path-file' to download specific directories / files, or '--get-all' to download all project contents."
         )
         sys.exit(1)
 
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -1607,7 +1607,7 @@ def get_data(
         sys.exit(1)
     elif not get_all and not (source or source_path_file):
         LOG.error(
-            "Specify either '--source'/'--source-path-file' to download specific directories / files, or '--get-all' to download all project contents."
+            "Specify either '--source'/'--source-path-file' to download specific directories/files, or '--get-all' to download all project contents."
         )
         sys.exit(1)
 
--- a/dds_cli/__main__.py
+++ b/dds_cli/__main__.py
@@ -1606,7 +1606,9 @@ def get_data(
         )
         sys.exit(1)
     elif not get_all and not (source or source_path_file):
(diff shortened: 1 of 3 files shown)


dds_cli/__main__.py
```diff
@@ -1605,6 +1605,11 @@ def get_data(
             "Flag '--get-all' cannot be used together with options '--source'/'--source-path-fail'."
         )
         sys.exit(1)
+    elif not get_all and not (source or source_path_file):
+        LOG.error(
+            "The command 'dds data get' reuires either the flag '--get-all' or some of the options '--source'/'--source-path-fail'."
```
inaod568, line 1610: Suggestion to change the message a little
```suggestion
            "Specify either '--source'/'--source-path-file' to download specific directories / files, or '--get-all' to download all project contents."
```

inaod568 (commented):
Nice! Have suggested a slight clarification in the message

dds_cli/__main__.py
```diff
@@ -1605,6 +1605,11 @@ def get_data(
             "Flag '--get-all' cannot be used together with options '--source'/'--source-path-fail'."
         )
         sys.exit(1)
+    elif not get_all and not (source or source_path_file):
+        LOG.error(
+            "Specify either '--source'/'--source-path-file' to download specific directories/files, or '--get-all' to download all project contents."
```
inaod568, line 1610: Just a thought, do you think this will be interpreted as that `--source` is for directories and `--source-path-file` is for files? Maybe in that case we should have "directories or files" or something

valyo: for me `directories/files` = `directories or files`, doesn't it?
It doesn't matter for me ... actually the `either` kind of requires `or`, like:

"Specify either '--source' or '--source-path-file' ... 

and then the second one can be either `/` or `or`

inaod568: Yeah it is but it kind of looked like the `--source/--source-path-file` was meant to match `directories/files`, or at least I could see how someone would interpret it like that. 

Yeah do the 
```
"Specify either '--source' or '--source-path-file' to download specific directories/files, or '--get-all' to download all project contents."
```

valyo: yes, `/` on both places is straightforward, but the `either` doesn't go with the slashes. I am changing it.

inaod568 (approved):
Approved!

PR #592, opened 2023-01-10
Publish to TestPyPi on PR and release 

i-oden:
> **Before submitting the PR**
>
> - Fill in and tick fields
> - _Remove all rows_ that are not relevant for the current PR
>   - Revelant option missing? Add it as an item and add a PR comment informing that the new option should be included into this template.
>

# Description

- [x] Summary of the changes and the related issue: 

In order to debug and prevent potential issues during releases, we should also 1. publish the CLI to TestPyPi on pr to dev and on release before real publishing, 2. make sure that the hashes are correct. Possible that we also should recommend verifying the hashes after installation. 

The `print_hash: true` part in tells the action to show the hashes of what will be published. Easily verified manually by checking on PyPi. If something happens though and there's an error, the publishing will not occur. 

- Fixes an issue in GitHub / Jira:
  - [x] Yes: DDS-1443

## Type of change

- [x] Workflow

# Checklist:

## General

## Repository / Releases

## Checks

- [x] CodeQL passes
- Tests
  - [x] The tests pass
- Trivy:
  - [x] There are no new security alerts

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -42,7 +42,6 @@ jobs:
           user: __token__
           password: ${{ secrets.pypi_password }}
           skip_existing: true
-      
 
   ddsclibinaries:
     name: Build binary packages for the DDS CLI
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -25,6 +25,15 @@ jobs:
       - name: Build the distribution
         run: python setup.py sdist bdist_wheel
 
+      - name: Publish dist to TestPyPI
+        if: github.repository == 'ScilifelabDataCentre/dds_cli'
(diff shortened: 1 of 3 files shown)


valyo (approved):
Looks good. 
I don't completely understand the hashes part but you can probably explain it to me additionally

i-oden:
> Looks good. I don't completely understand the hashes part but you can probably explain it to me additionally

Before publishing to PyPi the action generates a checksum. These hashes are then available on PyPi, in the "Download" section, in case someone wants to verify them. This "print_hashes" part basically just prints these hashes out here on GitHub so that we could technically check that these hashes are identical to the ones on PyPi and we could possibly add this information to the documentation somehow (not sure though), to recommend that all users verify the integrity of the package. As a precaution and threat mitigation.

PR #655, opened 2023-09-26
Dds 1401 print project info when deleting archiving project

rv0lt:
## Before submitting this PR

1. **Description:** When deleting or achiving a project a message is printed with the project information before asking for confirmation.

2. **Jira task / GitHub issue:** https://scilifelab.atlassian.net/jira/software/projects/DDS/boards/13?selectedIssue=DDS-1401
3. **How to test:** dds_cli project status archive -p project_1 and then you should be able to see the information of the project and a confirmation prompt
4. **Type of change:** [_Check the relevant boxes in the section below_](#what-type-of-changes-does-the-pr-contain)
5. **Add docstrings and comments to code**, _even if_ you personally think it's obvious.

## What _type of change(s)_ does the PR contain?

- [x] New feature
  - [ ] Breaking: _Please describe the reason for the break and how we can fix it._
  - [x] Non-breaking
- [ ] Bug fix
  - [ ] Breaking: _Please describe the reason for the break and how we can fix it._
  - [ ] Non-breaking
- [ ] Security Alert fix
- [ ] Documentation
- [ ] Tests **(only)**
- [ ] Workflow

## Checklist

- [Sprintlog](../SPRINTLOG.md)
  - [x] Added
  - [ ] Not needed (E.g. PR contains _only_ tests)
- Rebase / Update / Merge _from_ base branch (the branch from which the current is forked)
  - [ ] Done
  - [ ] Not needed
- Blocking PRs
  - [ ] Merged
  - [x] No blocking PRs
- PR to `master` branch
  - [ ] Yes: Read [the release instructions](../docs/procedures/new_release.md)
    - [ ] I have followed steps 1-7.
  - [x] No

## Actions / Scans

- **Black**: Python code formatter. Does not execute. Only tests.
  Run `black .` locally to execute formatting.
  - [x] Passed
- **Pylint**: Python code linter. Does not execute. Only tests.
  Fix code producing warnings. Code must get 10/10.
  - [ ] Warnings fixed
  - [x] Passed
- **Prettier**: General code formatter. Our use case: MD and yaml mainly.
  Run `npx prettier --write .` locally to execute formatting.
  - [x] Passed
- **Yamllint**: Linting of yaml files.
  - [x] Passed
- **Tests**: Pytest to verify that functionality works as expected.
  - [ ] New tests added
  - [x] No new tests
  - [x] Passed
- **TestPyPi**: Build CLI and publish to TestPyPi in order to verify before release.
  - [x] Passed
- **CodeQL**: Scan for security vulnerabilities, bugs, errors
  - [ ] New alerts: _Go through them and either fix, dismiss och ignore. Add reasoning in items below._
  - [ ] Alerts fixed: _What?_
  - [ ] Alerts ignored / dismissed: _Why?_
  - [x] Passed
- **Trivy**: Security scanner
  - [ ] New alerts: _Go through them and either fix, dismiss och ignore. Add reasoning in items below._
  - [ ] Alerts fixed: _What?_
  - [ ] Alerts ignored / dismissed: _Why?_
  - [x] Passed
- **Snyk**: Security scanner
  - [ ] New alerts: _Go through them and either fix, dismiss och ignore. Add reasoning in items below._
  - [ ] Alerts fixed: _What?_
  - [ ] Alerts ignored / dismissed: _Why?_
  - [x] Passed

### DIFF
--- a/dds_cli/base.py
+++ b/dds_cli/base.py
@@ -148,6 +148,11 @@ class DDSBaseClass:
         )
 
         project_info = response.get("project_info")
+
+        # If not project info was retrieved from the request, throw an exception
+        if not project_info:
+            raise dds_cli.exceptions.ApiResponseError(message="No project information to display.")
+
         return project_info
 
     def generate_project_table(self, project_info):
--- a/dds_cli/base.py
+++ b/dds_cli/base.py
@@ -133,7 +133,7 @@ class DDSBaseClass:
 
         return True
 
-    # Private methods ############################### Private methods #
+    # Public methods ############################### Public methods #
 
     def get_project_info(self):
         """Collect project information from API."""
--- a/dds_cli/base.py
+++ b/dds_cli/base.py
@@ -148,9 +148,6 @@ class DDSBaseClass:
         )
 
         project_info = response.get("project_info")
-        if not project_info:
-            raise dds_cli.exceptions.ApiResponseError(message="No project information to display.")
-
         return project_info
 
     def generate_project_table(self, project_info):
--- a/dds_cli/base.py
+++ b/dds_cli/base.py
@@ -134,6 +134,41 @@ class DDSBaseClass:
         return True
 
     # Private methods ############################### Private methods #
+
+    def get_project_info(self):
+        """Collect project information from API."""
+
+        # Get info about a project from API
+        response, _ = dds_cli.utils.perform_request(
+            DDSEndpoint.PROJ_INFO,
+            method="get",
+            headers=self.token,
+            params={"project": self.project},
+            error_message="Failed to get project information",
+        )
+
+        project_info = response.get("project_info")
+        if not project_info:
+            raise dds_cli.exceptions.ApiResponseError(message="No project information to display.")
+
+        return project_info
+
+    def generate_project_table(self, project_info):
+        """Generate a table from some project info provided"""
+
+        # Print project info table
+        table = dds_cli.utils.create_table(
+            title="Project information.",
+            columns=["Project ID", "Created by", "Status", "Last updated", "Size"],
+            rows=[
+                project_info,
+            ],
+            caption=f"Information about project {project_info['Project ID']}",
+        )
+
+        return table
+
+    # Private methods ############################### Private methods #
     def __get_safespring_keys(self):
         """Get safespring keys."""
         return s3.S3Connector(project_id=self.project, token=self.token)
--- a/dds_cli/project_info.py
+++ b/dds_cli/project_info.py
@@ -44,37 +44,13 @@ class ProjectInfoManager(base.DDSBaseClass):
         self.project = project
 
     # Public methods ###################### Public methods #
-    def get_project_info(self):
-        """Collect project information from API."""
-        # Get info about a project from API
-        response, _ = dds_cli.utils.perform_request(
-            DDSEndpoint.PROJ_INFO,
-            method="get",
-            headers=self.token,
-            params={"project": self.project},
-            error_message="Failed to get project information",
-        )
-
-        project_info = response.get("project_info")
-        if not project_info:
-            raise dds_cli.exceptions.ApiResponseError(message="No project information to display.")
-
-        return project_info
 
     def show_project_info(self):
         """Display info about a project."""
-        # Get project info from API
-        project_info = self.get_project_info()
 
-        # Print project info table
-        table = dds_cli.utils.create_table(
-            title="Project information.",
-            columns=["Project ID", "Created by", "Status", "Last updated", "Size"],
-            rows=[
-                project_info,
-            ],
-            caption=f"Information about project {project_info['Project ID']}",
-        )
+        # Get project info from api
+        project_info = self.get_project_info()
+        table = self.generate_project_table(project_info=project_info)
         dds_cli.utils.console.print(table)
 
         # Print Title and Description below the table
--- a/dds_cli/project_status.py
+++ b/dds_cli/project_status.py
@@ -124,7 +124,6 @@ class ProjectStatusManager(base.DDSBaseClass):
 
         # If the status is going to be archived or deleted. Ask for confirmation
         if new_status in ["Archived", "Deleted"]:
-
             # get project info
             project_info = self.get_project_info()
 
--- a/dds_cli/project_status.py
+++ b/dds_cli/project_status.py
@@ -52,6 +52,7 @@ class ProjectStatusManager(base.DDSBaseClass):
     def get_project_info(self):
         """Collect project information from API."""
         # Get info about a project from API
+
         response, _ = dds_cli.utils.perform_request(
             DDSEndpoint.PROJ_INFO,
             method="get",
@@ -61,9 +62,6 @@ class ProjectStatusManager(base.DDSBaseClass):
         )
 
         project_info = response.get("project_info")
-        if not project_info:
-            raise dds_cli.exceptions.ApiResponseError(message="No project information to display.")
-
         return project_info
 
     def get_status(self, show_history):
--- a/dds_cli/project_status.py
+++ b/dds_cli/project_status.py
@@ -114,7 +114,7 @@ class ProjectStatusManager(base.DDSBaseClass):
                 project_info = self.get_project_info()
             except exceptions.ApiResponseError:
                 dds_cli.utils.console.print(
-                    f"No project information could be displayed at this moment!"
+                    "No project information could be displayed at this moment!"
                 )
             else:
                 table = self.generate_project_table(project_info=project_info)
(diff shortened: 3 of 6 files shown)


dds_cli/project_status.py
```diff
@@ -102,6 +122,31 @@ def update_status(self, new_status, deadline=None, is_aborted=False, no_mail=Fal
...
+                f"Description: \t{project_info['Description']}\n"
+                f"PI: \t{project_info['PI']}\n"
+            )
+
+            dds_cli.utils.console.print(print_info)
+
+            if not rich.prompt.Confirm.ask("-"):
+                LOG.info("Probably for the best. Exiting.")
+                sys.exit(0)
+
```
i-oden, line 149: I like it, but would like to display info first and then ask for confirmation
It would be nice to have a table though. The function `show_project_info` in `project_info.py` both calls `get_project_info()` and then prints the project information as a table. It would be good if we could reuse this function, but then we'd either need to change it to a static method, or use the `ProjectInfoManager` class, or move the code to a helper function that both the `ProjectInfoManager` and `ProjectStatusManager` can use. 
   
```suggestion
        # If the status is going to be archived or deleted. Ask for confirmation
        if new_status in ["Archived", "Deleted"]:
            # Create confirmation prompt with project info
            dds_cli.utils.console.print(f"The following project '[b]{self.project}[/b]' is about to be [b][blue]{new_status}[/blue][/b].\n")
            
            # Display project info
            dds_cli.project_info.show_project_info() # needs some more because it's a public method in the ProjectInfoManager
            
            # Define question
            question = f"Are you sure you want to modify the status of this project? All its contents "
            if new_status == "Deleted":
                print_info += "and metainfo "
            print_info += "will be deleted!\n"
            
            # Ask for confirmation 
            if not rich.prompt.Confirm.ask(question):
                LOG.info("Probably for the best. Exiting.")
                sys.exit(0)
```

i-oden (changes requested):
Suggestion to reuse existing code, but would need some tweeking

dds_cli/project_status.py
rv0lt: @i-oden _would like to display info first and then ask for confirmation_
 What do you mean by that? It already prints the info and then asks for the user input.
 I can look at the table thing. But wouldn't it make it more difficult to test the output?

i-oden: Yeah I was a bit unclear here. At the moment it prints the following: 
```bash
Are you sure you want to modify the status of someunit00001? All its contents and metainfo will be deleted!
The project 'someunit00001' is about to be Deleted.
Title:  test
Description:    test
PI:     test@mail.com

- [y/n]: 
```
So 
1. Question
2. Info
3. Ask for input

But I'd like it to be 
1. Info
2. Question and ask for input 
----

Regarding testing - yes, the tables make it more difficult. But there are tests that check tables, e.g. `test_list_all_active_motds_table`, so check that one out for example. If testing the actual table output is too complicated and just not worth it: Skip the testing of exactly that part and only test the rest of the output etc.

i-oden: What you've done is really good, my suggestions are just some slight improvements, partly in order to reduce code reuse, and partly due to previous experience with the users. It needs to be super clear and there needs to be clear separate sections in the print outs. So
1. 'You're trying to do this'
2. <project information>
3. 'Are you sure, this will happen' 
will make it clearer for the unit users.
So in this case the tests are not the most important part, here's it's the user experience etc. The tests are a bonus but not a must.

rv0lt:
@i-oden Moved the _get project info_ and the _generation of the table_ to the base class. So both of them can inherit that function.

The new output is like:

`/code # dds project status delete -p project_1
     ︵ 
 ︵ (  )   ︵ 
(  ) ) (  (  )   SciLifeLab Data Delivery System 
 ︶  (  ) ) (    http://dds_backend:5000/ 
      ︶ (  )    CLI Version 2.5.1 
          ︶
Current user: unitadmin_1
                        Project information.                         
┏━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━┓
┃ Project ID ┃ Created by      ┃ Status      ┃ Last updated ┃ Size  ┃
┡━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━┩
│ project_1  │ First Unit User │ In Progress │ None         │ 0.0 B │
└────────────┴─────────────────┴─────────────┴──────────────┴───────┘
                 Information about project project_1                 
Are you sure you want to modify the status of project_1? All its contents and metainfo will be deleted!
The project project_1 is about to be Deleted.

-[y/n]: n
INFO     Probably for the best. Exiting.                                                                                                                                                                                                                                                                                 `


The tests I was also able to find a easy way to check them.

dds_cli/base.py
```diff
@@ -133,6 +133,38 @@ def __exit__(self, exception_type, exception_value, traceback, max_fileerrs: int
 
         return True
 
+    # Private methods ############################### Private methods #
```
i-oden, line 136:
```suggestion
    # Public methods ############################### Public methods #
```

dds_cli/project_status.py
```diff
@@ -102,6 +107,29 @@ def update_status(self, new_status, deadline=None, is_aborted=False, no_mail=Fal
...
+            print_info = (
+                f"Are you sure you want to modify the status of {self.project}? All its contents "
+            )
+            if new_status == "Deleted":
+                print_info += "and metainfo "
+            print_info += (
+                "will be deleted!\n" f"The project {self.project} is about to be {new_status}.\n"
+            )
+
+            dds_cli.utils.console.print(print_info)
```
i-oden, line 127: I liked how you had colored certain things before my last review - could you add so that
- `{self.project}` is `'{self.project}'`
- `{new_status}` is bold and blue like you had it before

This also helped. Other than that - this looks great.

dds_cli/project_status.py
```diff
@@ -102,6 +107,29 @@ def update_status(self, new_status, deadline=None, is_aborted=False, no_mail=Fal
...
+
+            if not rich.prompt.Confirm.ask("-"):
+                LOG.info("Probably for the best. Exiting.")
+                sys.exit(0)
+
```
i-oden, line 132: Technically you could just put `print_info` into the `rich.prompt.Confirm.ask`, but this is just a very minor thing so decide for yourself if you want to do that or not

dds_cli/project_info.py
```diff
@@ -44,37 +44,13 @@ def __init__(
...
-            DDSEndpoint.PROJ_INFO,
-            method="get",
-            headers=self.token,
-            params={"project": self.project},
-            error_message="Failed to get project information",
-        )
-
-        project_info = response.get("project_info")
-        if not project_info:
-            raise dds_cli.exceptions.ApiResponseError(message="No project information to display.")
```
i-oden, line 60: Atm if "project_info" is not returned for some reason, a traceback is printed. I think we should add this again to the `get_project_info` method in `base.py`. In order to let the users delete or archive the project even if the project info is not displayed, we can add a `try-except` in `update_status` around line 113, and the `except` can print out for example "Could not display project information at this time." and then move on to the confirmation, exactly like you have it now. 

Thoughts / other suggestions on that?

i-oden: **Additional comment that has nothing to do with this PR - don't do this in this PR, it's outside of the scope:**
Future task: Catch the exception raised in the `get_project_info` method where it's raised, if we don't want it to completely quit.
- `show_project_info`, line 52: Should **not** catch the exception, it should still be raised
- `update_info`, line 70: Should catch the exception with a try-except. The except could for example print and ask for confirmation from the user: "Could not display project information. Are you sure you want to change the project info?"

i-oden (changes requested):
Looks good and the steps of the printouts feel more natural. Also nicely done with the tests
I have a few questions and suggestions, partly regarding an exception that was removed, but also that I liked how you had it before with the coloring etc, so please add that again.

dds_cli/project_info.py
rv0lt: I removed it mainly because the requests was already throwing an exception, so, for example, when the request fails it already leaves the function. The coverage was never getting to 100% because of that as well. The _if not project_info_: was never being executed. I can add another try-catch in this method if we want to generalise every problem in this requests through the _ApiResponseError_

dds_cli/project_status.py
rv0lt: I like leaving that being the same ProjectInfo and ProjectStatus, more generalized

dds_cli/project_info.py
i-oden: Yeah I see what you mean, this check is more to verify that "project_info" actually is returned, so that if the API returns something, but not the term "project_info", then it prints out a message that we can easily interpret instead of a traceback that might also result in support tickets etc. So it's good to keep this check anyway

i-oden (approved):
Looks great! 
Also I like the monkeypatch thing in the tests.
You can merge

PR #737, opened 2025-03-05
DOESN'T WORK -- Log file created by default for uploads and downloads

i-oden:
## Read this before submitting the PR

1. Always create a Draft PR first
2. Go through sections 1-5 below, fill them in and check all the boxes
3. Make sure that the branch is updated; if there's an "Update branch" button at the bottom of the PR, rebase or update branch.
4. When all boxes are checked, information is filled in, and the branch is updated: mark as Ready For Review and tag reviewers (top right)
5. Once there is a submitted review, implement the suggestions (if reasonable, otherwise discuss) and request an new review.

If there is a field which you are unsure about, enter the edit mode of this description or go to the [PR template](../.github/pull_request_template.md); There are invisible comments providing descriptions which may be of help.

## 1. Description / Summary

People rarely use the `--log-file` option. Previously the default was that there was always a log file generated, but we were logging a lot, which resulted in the users (mainly the units) generating a huge amount of logs which were quite big. However, we have now reduced the logging. 

Every time there's a support ticket, we usually end up having to ask the user to run the command again, which takes both time and energy for us, but also of course means that we cannot immediately investigate what the issue is and often have to wait for the user to try again. 

This PR adds a flag `--force-no-log` which the user can use if they really do not want to generate a log file. If this flag is not used, the dds by default saves all logs to a file with the format: `<command>_<current time>.log`. The user can also use the `--log-file` option to specify their own path to the log file, as usual. 

This PR also creates a directory in `home` called `dds-cli_logs` where the automatic default logs will be stored. I could technically have used the generated DDS directory which is always created when uploads and downloads are started but this meant more code changes in more files, and made everything less maintainable.

## 2. Jira task / GitHub issue

**Is this a GitHub issue?** --> Add the link to the github issue

**Is this from a Jira task?** --> If your branch does not contain info regarding the Jira task ID, put it here.

## 3. Type of change - Add label

**What _type of change(s)_ does the PR contain? For an explanation of the different options below, enter edit mode of this PR description template.**

_If you do not want this change to be included in release notes, add the label `skip-changelog`._

- New feature
  - Breaking --> label: `breaking` 
  - Non-breaking --> label: `feature` 
- Database change --> label: `feature` or none at all.
  _Remember the to include a new migration version, **or** explain here why it's not needed._ 
- Bug fix --> label: `bug` 
- Security Alert fix 
  - Package update --> label: `dependency` 
    - Major version update 
- Documentation --> label can be skipped, will be included in "other changes" 
[...]

PR #740, opened 2025-03-27
Log file created by default for uploads and downloads

i-oden:
## Read this before submitting the PR

1. Always create a Draft PR first
2. Go through sections 1-5 below, fill them in and check all the boxes
3. Make sure that the branch is updated; if there's an "Update branch" button at the bottom of the PR, rebase or update branch.
4. When all boxes are checked, information is filled in, and the branch is updated: mark as Ready For Review and tag reviewers (top right)
5. Once there is a submitted review, implement the suggestions (if reasonable, otherwise discuss) and request an new review.

If there is a field which you are unsure about, enter the edit mode of this description or go to the [PR template](../.github/pull_request_template.md); There are invisible comments providing descriptions which may be of help.

## 1. Description / Summary

People rarely use the --log-file option. Previously the default was that there was always a log file generated, but we were logging a lot, which resulted in the users (mainly the units) generating a huge amount of logs which were quite big. However, we have now reduced the logging a bit.

Every time there's a support ticket, we usually end up having to ask the user to run the command again, which takes both time and energy for us, but also of course means that we cannot immediately investigate what the issue is and often have to wait for the user to try again.

This PR adds a flag --force-no-log which the user can use if they really do not want to generate a log file. If this flag is not used, the dds by default saves all logs to a file with the format: <command>_<current time>.log. The user can also use the --log-file option to specify their own path to the log file, as usual.

The code needed some refactoring before being able to do this, and at first I made this PR: https://github.com/ScilifelabDataCentre/dds_cli/pull/737
However, while that did technically work when it comes to generating the actual log file as I wanted it to, it did not actually log things to the file. In order to make it work I needed to have it in the `__main__.py` module. There is a slight bit of code duplication since I need to add code to the `dds data get` and `dds data put` commands. 

I probably need to look at adding tests as well, but will look at that next week. 

## 2. Jira task / GitHub issue

HMS-2292

## 3. Type of change - Add label

**What _type of change(s)_ does the PR contain? For an explanation of the different options below, enter edit mode of this PR description template.**

_If you do not want this change to be included in release notes, add the label `skip-changelog`._

- New feature
  - Breaking --> label: `breaking` 
  - Non-breaking --> label: `feature` 
- Database change --> label: `feature` or none at all.
  _Remember the to include a new migration version, **or** explain here why it's not needed._ 
- Bug fix --> label: `bug` 
- Security Alert fix 
  - Package update --> label: `dependency` 
    - Major version update 
- Documentation --> label can be skipped, will be included in "other changes" 
- Workflow --> label: `skip-changelog` 
- Tests **only** --> label: `skip-changelog` 

## 4. Additional information

- [x] I have added an entry to the [Sprintlog](../SPRINTLOG.md) 
- [ ] This is a PR to the `master` branch: _If checked, read [the release instructions](../doc/procedures/new_release.md)_ 
  - [ ] I have followed steps 1-8. 

## 5. Actions / Scans

**Make sure that the following checks/actions have passed.**

- **Black**

- **Prettier**

- **Pylint**

- **Yamllint**

- **Tests**

- **CodeQL**

- **Trivy**

- **Snyk**

- **TestPyPI**

If an action does not pass and you need help with how to solve it, enter edit mode of this PR template or go to the [PR template](../.github/pull_request_template.md).


i-oden:
P.s. I will rename this PR before we merge it, I just want to keep track of which one is the working one and which one I made first etc.

i-oden:
@ScilifelabDataCentre/teamhermes Review this when you have a chance

dds_cli/directory.py
```diff
@@ -29,16 +29,23 @@
 class DDSDirectory:
     """Data Delivery System directory class."""
 
-    def __init__(self, path=pathlib.Path, add_file_dir: bool = True):
+    def __init__(
+        self,
+        path=pathlib.Path,
+        add_file_dir: bool = True,
```
valyo, line 35: Just wondering what the purpose of `add_file_dir` is, since I don't see it being used anywhere in the code

valyo (changes requested):

dds_cli/__main__.py
```diff
@@ -153,20 +173,18 @@ def dds_main(click_ctx, verbose, log_file, no_prompt, token_path):
...
-        # Create context object
-        click_ctx.obj = {"NO_PROMPT": no_prompt, "TOKEN_PATH": token_path}
+            if force_no_log:
+                LOG.warning(
+                    "You have used both the '--log-file' and '--force-no-log' option, you can only use one."
+                )
+                sys.exit(1)
+            else:
+                file_handler = dds_cli.utils.setup_logging_to_file(filename=log_file)
+                LOG.addHandler(file_handler)
```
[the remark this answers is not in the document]
i-oden: Fixed

dds_cli/directory.py
i-oden: It's used on line below.

valyo: I meant that it doesn't seem to be set anywhere and passed to the class init, so it always defaults to True. Or am I missing something?

rv0lt:
Maybe the only detail that I see. Is that when there is an error, the message asks to check for the JSON log and not this new one; example:

`Please verify that the following error log has been generated: /code/DataDelivery_2025-04-25_08-49-38_project_1_upload/logs/dds_failed_delivery.json`

i-oden:
> Maybe the only detail that I see. Is that when there is an error, the message asks to check for the JSON log and not this new one; example:
> 
> `Please verify that the following error log has been generated: /code/DataDelivery_2025-04-25_08-49-38_project_1_upload/logs/dds_failed_delivery.json`

Good point!

rv0lt (approved):

i-oden:
@rv0lt I added the same info to the download message

PR #799, opened 2025-08-22
Use explicit home token path

i-oden:
## Summary
- Ensure GUI state manager uses the user's home directory for the CLI token file

------
https://chatgpt.com/codex/tasks/task_b_68a85af681b48326b93c41e5f76037ac

### DIFF
--- a/dds_cli/dds_gui/dds_state_manager.py
+++ b/dds_cli/dds_gui/dds_state_manager.py
@@ -1,7 +1,7 @@
 """DDS State Manager"""
 
 from dataclasses import dataclass
-from pathlib import Path
+import pathlib
 
 from textual.reactive import reactive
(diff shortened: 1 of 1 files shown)


ebbaoliveberg (commented):
In the description it says:

- python -m py_compile dds_cli/dds_gui/dds_state_manager.py
- pytest -q (fails: ModuleNotFoundError: No module named 'pyfakefs')
- pip install pyfakefs requests_mock flask -q (fails: Could not find a version that satisfies the requirement pyfakefs)

This is a bit confusing, for me it woks locally and I don't think we test anything in that manner elsewhere. What is the reasoning of that part of the description?

ebbaoliveberg (approved):

